PR #37, opened 2011-07-09
Reduce the number of items in the action bar when using tab navigation

ChrisSmith:
Currently the items in the action bar take up too much space when using tab navigation. Landscape is ok, but in portrait the tabs become to small.

This commit reduces the available spaces to 1 while using tab navigation in portrait.

### DIFF
--- a/library/src/com/actionbarsherlock/internal/app/ActionBarSupportImpl.java
+++ b/library/src/com/actionbarsherlock/internal/app/ActionBarSupportImpl.java
@@ -54,6 +54,9 @@ public final class ActionBarSupportImpl extends ActionBar {
 	
 	/** Maximum action bar items in portrait mode. */
 	private static final int MAX_ACTION_BAR_ITEMS_PORTRAIT = 3;
+
+	/** Maximum action bar items in portrait mode with tab navigation. */
+	private static final int MAX_ACTION_BAR_ITEMS_PORTRAIT_TABS = 1;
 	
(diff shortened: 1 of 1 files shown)


JakeWharton:
I am incredibly torn on things like this. I definitely agree that tabs and action items in portrait are a crowded affair. However, I'm hesitant to hard-code a limit the action items to only one when in tab navigation mode. I do like the idea of limiting how many items can be displayed (to an extent). I also think that perhaps this is on the responsibility of the implementer on limiting how many items they add, knowing what navigation type they are currently using. Using a `menu` and `menu-v11` version of their XML or setting the `showAsAction` value programmatically by checking `Build.VERSION.SDK_INT` allows this manual control with fairly minimal effort.

I'm going to be doing a lot of work with the action item inflation tomorrow so let me think on it a bit.

ChrisSmith:
Relying on `menu` and `menu-v11` is not enough to get the desired results, you would also have to include qualifiers for `port` and `land`, which is far to much duplication in my option.  Doing this in code at inflation is also too much because you have the same checks in every file, not very DRY. 

I understand not wanting to hard code the limit, because sometimes it doesn't apply and the user may want two items instead of one.  So instead of a hard coded limit, what if that limit was exposed via a setter method? Something like `setMaxActionItemsPort(int max)` and `setMaxActionItemsLand(int max)`. At least this way there is some flexibility and it can be used regardless of navigation mode.

PR #51, opened 2011-07-30
Use IFragementActivity for all local references

felixb:
Let FragmentActivity implement the introduced IFragmentActivity.
Use interface for all references.

This will let you add a local FragmentMapActivity by code duplication of FragmentActivity without the need of letting all Activities inherent MapActivity.

### DIFF
--- a/library/src/android/support/v4/app/ActionBar.java
+++ b/library/src/android/support/v4/app/ActionBar.java
@@ -34,10 +34,10 @@ import com.actionbarsherlock.internal.view.menu.MenuItemImpl;
  */
 public abstract class ActionBar {
 	/** Parent activity. */
-	private final FragmentActivity mActivity;
+	private final IFragmentActivity mActivity;
 	
 	
-	protected ActionBar(FragmentActivity activity) {
+	protected ActionBar(IFragmentActivity activity) {
 		mActivity = activity;
 	}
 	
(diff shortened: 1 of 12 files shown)


JakeWharton:
Thank you for this. I have been referring people to the [Google Group posting](https://groups.google.com/d/msg/android-developers/twW71y0o0AA/iqCS08cBSxoJ) when they'd email and didn't want to change `FragmentActivity` to just extend `MapActivity`. This won't make it into 3.1.1 but I'll consider it for 3.1.2 or 3.2.0.

I was always a bit wary of this solution since it requires exposing some internals of `FragmentActivity` but there really is no alternative until someone (hopefully Google) release a true MapView that works like the V3 JavaScript implementation.

JakeWharton:
Haven't forgotten about this. Going to release 3.1.3 and then try and come with a solution that is the least common denominator of suck to support maps. I don't like exposing the private, behind-the-scenes APIs to users of the library. I don't want to have to resort to reflection, however. Maybe two interfaces: one public, one package scoped?

Just need a damn `MapFragment` already...

JakeWharton:
This has been implemented in the `dev` branch using a similar method. Feel free to test and report back your thoughts.

felixb:
looks good to me.
here is my FragmentMapActivity:
https://github.com/felixb/travellog/blob/master/src/android/support/v4/app/FragmentMapActivity.java

JakeWharton:
Released with 3.2.0

PR #52, opened 2011-08-08
Visual error when hiding ActionBar.

xperimental:
Hi,

I'm wondering why you are setting the ActionBar custom background (set using `abBackground` attribute) to both the `ActionBarContainer` and `ActionBarView`.

For me this causes a small visual error when hiding the action bar on pre-3.0 devices, as only the `ActionBarView` is hidden not the `ActionBarContainer`. I'm using a nine-patch as background image and the "non-content" part of the background image stays visible.

Removing the background image from the `ActionBarContainer` (see commit) has resolved this issue for me.

Regards

### DIFF
--- a/library/src/com/actionbarsherlock/internal/widget/ActionBarContainer.java
+++ b/library/src/com/actionbarsherlock/internal/widget/ActionBarContainer.java
@@ -16,10 +16,6 @@ public class ActionBarContainer extends FrameLayout {
     }
     public ActionBarContainer(Context context, AttributeSet attrs) {
         super(context, attrs);
-
-        TypedArray a = context.obtainStyledAttributes(attrs, R.styleable.SherlockTheme);
-        setBackgroundDrawable(a.getDrawable(R.styleable.SherlockTheme_abBackground));
-        a.recycle();
     }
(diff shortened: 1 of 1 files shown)


JakeWharton:
Can you provide me with the 9-patch to test with? You're on the right track except from looking at decompiled Android 3.2 code the background is explicitly removed from the `ActionBarView` rather than the container.

Have you tried this on Android 3.0+?

xperimental:
I do not have the nine-patch available right now, but I can send it to you tomorrow if you need it. It's a very simple one, 5x5 pixel in size, ASCII-art follows:

```
T T X T T
T B B B T
X B B B X
T Y Y Y T
T T X T T

T = transparent
X = black
B = blue
Y = yellow
```

When the action bar was hidden two lines remained on screen. One blue and one yellow, so I suppose that it was the two "non-content" lines from the nine-patch.

Also, I did test it on a 3.0.1 Motorola Xoom. It works without problems there.

JakeWharton:
That's perfect. Will test tonight and (hopefully) push a fix to `3.1.x` branch.

PR #75, opened 2011-10-11
Do not call onActivityResult on null fragment.

xperimental:
Hi,

I think I've found a small bug in the library. As you can see the `FragmentActivity` checks in `onActivityResult` if the requesting fragment still exists and logs a warning if the reference is null. But it still runs into a NPE if the `Fragment` is null because the call to `Fragment.onActivityResult` is always done.

Is this the expected behaviour?

Regards
  Robert


JakeWharton:
Nice find. Looking to release either tonight or this weekend after some final testing and tweaking.

Thanks!

PR #76, opened 2011-10-16
Fix DialogFragment dismissing on orientation change.

SimonVT:
mShowsDialog indicates whether a DialogFragment is shown as a dialog. It's initiated as false, and changed to true if .show(...) is called on the DialogFragment.
However, on orientation change it is only saved if it's false, causing the dialog to not be shown, or crashing if getDialog() is called any time during creation.


JakeWharton:
Relates to 5c315d6140c39c61aa39f96bca10432e957293c8.

PR #77, opened 2011-10-17
Menu resource onClick attribute is broken (refers to wrong object)

mernen:
`MenuInflater` attempts to bind the `onClick` handlers to itself, rather than to the given Context. This patch fixes this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Version 3.3.1 *(In Development)*
 
  * XML-defined `onClick` attributes will now check for an `onClick` method that
    takes an `android.support.v4.view.MenuItem` instance.
+ * Fix: Menu inflater properly checks activity context for `onClick` method
+   declared in the XML.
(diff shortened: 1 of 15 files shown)


mernen:
Also, perhaps it'd be a good idea to support handlers that take an `android.support.v4.view.MenuItem` (rather than `android.view.MenuItem`) as a parameter? I was quite surprised when it said my method was not found, I just assumed the library would favor its backport variants at all times over OS-provided classes.

JakeWharton:
Good catch. I always use in-code bindings so this slipped through.

I will also change the callback to use the support version of `MenuItem`.

Thanks!

mernen:
I wanted to give them a spin, but I'm a bit underwhelmed. At the moment they are limited to Context/Activity methods, and sometimes (quite often, actually) it makes much more sense to handle actions in their originating Fragment.

JakeWharton:
Merged to `dev`. Release soon.

PR #81, opened 2011-10-19
Fix NPE when calling ActionBarImpl.getNavigationItemCount() when tab navigation is enabled and no tab is actually selected. Return -1 in this case, which is what the Honeycomb implementation does.

zachlipton:
Returning -1 is undocumented behavior, but empirical observation reveals this to be a part of Google's implementation.


JakeWharton:
Merged to `dev` @ 5fd238aa31b21370cd979f7e7def4a20031aac70.

PR #82, opened 2011-10-19
Fix ArrayIndexOutOfBoundsException when removing a menu group

zachlipton:
Fix ArrayIndexOutOfBoundsException when removing a menu group from the overflow menu.

After removing an item, decrement the loop iterator so we don't walk off the end of the array.


JakeWharton:
Merged to `dev` @ 95f49c75955de3aff708fef390e761f5305a22ec.

PR #83, opened 2011-10-19
Fix two NPEs in ActionBarWrapper

zachlipton:
Fix two NPEs in ActionBarWrapper that occured when calling getSelectedTab or getTabAt when no tab was selected or no tab was at the specified index. Added null checks. Honeycomb only.

This makes the behavior in these cases be the same as Android 3.2's implementation.


JakeWharton:
Merged to `dev` @ 6f777183d6383a4fd1fbe7288ea4133aa979f2be.

PR #88, opened 2011-10-21
Home custom layout

aectann:
Adds support for abHomeLayout stylebale.


aectann:
The support for the styleable is declared in theming http://actionbarsherlock.com/theming.html docs, but looks like it was never implemented.

JakeWharton:
Landed in commit 6f43855a387059171c4a23fdec8138b1ebdc663f.

JakeWharton:
It's actually in 59d837334a082f90e165beccf77778ad137351da. Stupid merging script automatically applied it to master...

PR #97, opened 2011-11-01
Change for home logo clipping

hameno:
If the home logo has bigger dimensions than expected it would be clipped.


JakeWharton:
Merged into `dev` branch.

PR #98, opened 2011-11-01
Tab Bar Enhancements + ViewPager helper method

chrisbanes:
Couple of things in this Pull.
- Inline Tabs are now shown in a HorizontalScrollView. Which mirrors what stock HC/ICS does.
- Improve when Inline Tabs are shown.
- Enhancements to ViewPager. Can now get access to the currently shown Object (View, Fragment, etc)
- Make makeFragmentName Public. It's useful when you need to get access to the Fragments created in FragmentPagerAdapter


JakeWharton:
My only _potential_ issue with this is that you are now exposing a new API on `ViewPager` which is not present in the actual compatibility library. I don't mind including bug fixes in the compat lib classes but I have to be wary of adding methods. I'm not wholly against it, I just need to consider it carefully.

Is there any reason that you are taking this approach rather than keeping track in an adapter via the `setPrimaryItem` callback?

chrisbanes:
I've only just seen `setPrimaryItem`. I'll have a look. 

The Tabs stuff is good though.

JakeWharton:
That was a recent addition with the r4 ACL. GIve it a try and let me know what you think. I'm going to merge this leaving that out for now.

And yeah the tabs and publicizing of the fragment name method (a common question in IRC) looks great.

chrisbanes:
`setPrimaryItem` doesn't really help with this. I've done this instead using the new `getCurrentItem` method in ViewPager and the up-scoped `makeFragmentName`.

Thanks

SimonVT:
This should really be done by using the already existing OnPageChangeListener rather than exposing new API.

chrisbanes:
My requirement was to get the Object (View, Fragment, etc) not the position int, that's already available in the new ACL library (`getCurrentItem`).

SimonVT:
Sorry, I mistook getCurrentItem() for a change made by you/Jake.

PR #100, opened 2011-11-02
Animate ActionBar show/hide

chrisbanes:


chrisbanes:
Good point.

shashidhar502:
How to avoid actionbar animation while hiding or showing ?

onecoders:
How to remove the Animation ActionBar show/hide?
When actionbar.hide() is called, the ActionBarWrapper's hide() is called, not ActionBarImpl's.
How to solve this, please give me help.
Thanks.

PR #103, opened 2011-11-04
This would be very useful 

iNoles:
It can help to reduce duplicate coding when you have multiple Activities.


JakeWharton:
Abstracting the action bar logic into its own class is something that I do want to move towards but with the 4.x version of ActionBarSherlock.

Some of the choices that you have made, however, I take a bit of issue with. The exposing of internal functionality such as the fragment manager and internal handler as public will inevitably lead to problems with users accessing them. This was the reason for the public internal callbacks method which had package-scoped methods.

I'll leave this open as it has some things which I do want to move toward.

iNoles:
Oh okay, looks like I need to find another way of doing it. 

getLoaderManager and ensureSupportActionBarAttached() right?

JakeWharton:
All of the methods that were in the internal callbacks shouldn't be public since they expose the behind-the-scenes implementations of certain fragile things.

iNoles:
Okay, I put back old internal callback method.

JakeWharton:
I'm closing this, if you look that the `4.0-wip` progress I am already using a centralized class for management.

PR #136, opened 2011-12-27
NPE fix.

yanchenko:
I've taken com.actionbarsherlock.sample.demos.app.ActionBarTabsPager as an example.
An app with 2 tabs, each contains it's own Fragment. A NPE occurs when launching an Activity from one of the Fragments.


JakeWharton:
This was already fixed in the `dev` branch as d023ad37433735b0a39f35c366669779765dc566 and will be released with v3.5.1 hopefully soon. Thanks for contributing though!

PR #152, opened 2012-01-03
Fixes resources not found exception when menu item has no icon

SimonVT:
^


JakeWharton:
Thanks. Turns out I forgot to merge the ICS `MenuInflater` altogether.

JakeWharton:
Released in beta 4 @ http://beta.abs.io/

PR #161, opened 2012-01-03
NullPointerException in FragmentManager.saveFragmentBasicState(...)

jondwillis:
FragmentManager: prevented NullPointerException when saving some types of Fragments where mSavedViewState is null and mUserVisibleHint is false.

Instances of this class were causing the NPE:

public abstract class LayoutFragment extends Fragment {

```
public static LayoutFragment newInstance(final int layoutId) {
    return new LayoutFragment() {
        @Override
        public int getLayoutResourceId() {
            return layoutId;
        }
    };
}

public abstract int getLayoutResourceId();

@Override
public View onCreateView(LayoutInflater inflater, ViewGroup container,
        Bundle savedInstanceState) {
    return inflater.inflate(getLayoutResourceId(), container, false);
}
```

}

### DIFF
--- a/library/src/android/support/v4/app/FragmentManager.java
+++ b/library/src/android/support/v4/app/FragmentManager.java
@@ -1572,6 +1572,9 @@ final class FragmentManagerImpl extends FragmentManager {
                     FragmentManagerImpl.VIEW_STATE_TAG, f.mSavedViewState);
         }
         if (!f.mUserVisibleHint) {
+            if (result == null) {
+                result = new Bundle();
+            }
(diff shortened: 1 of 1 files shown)


JakeWharton:
This is already fixed on the `dev` branch but I really appreciate the pull request. I've been sitting on it for about two weeks. I _swear_ I'll release 3.5.1 tonight.

jondwillis:
Ah, that might explain why it showed up today for me. I might've been compiling against your dev branch prior to today. Thanks and keep up the good word! Can't wait to see 4.0 finished.

PR #178, opened 2012-01-12
Made tab bar height configurable.

yanchenko:


JakeWharton:
The only problem with this is that it won't be reflected on Android 3.0+. I'm not sure how that hard coded value slipped through but tabs _should_ be the same height as the action bar.

yanchenko:
Agreed concerning the same height.
_library/res/layout/abs__action_bar.xml_ should then have

```
82 - android:layout_height="44dp"
82 + android:layout_height="?attr/abHeight"
```

PR #182, opened 2012-01-14
Minor changes to compatibility actionbar

billy1380:
- changed portrait version of compatibility action bar so that it sits
  in a horizontal scroll view - this looked really weird when there were
  too many tabs especially that all the auto scrolling text was moving
  that the same time - requires more that 4 tabs to test
- changed action bar view so that when a tab is selected (without user
  input) it scrolls into view if it was not visible - this matches
  behaviour in ICS (not sure about HC)

### DIFF
--- a/library/res/layout/abs__action_bar.xml
+++ b/library/res/layout/abs__action_bar.xml
@@ -76,16 +76,25 @@
 				android:visibility="gone" />
 		</RelativeLayout>
 	</RelativeLayout>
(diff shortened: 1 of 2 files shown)


JakeWharton:
This is covered under v4.0 which will have a release candidate tomorrow. Thanks for the pull request though!

billy1380:
Sound great... looking forward to it... Keep up the amazing work.

PR #194, opened 2012-01-17
4.0 wip

cuu508:
UI/Application Exerciser Monkey crashes my app in this place with NullPointerException. Adding this check fixed the problem.


JakeWharton:
Yep, this could probably happen. Thanks!

PR #197, opened 2012-01-22
Tabs same height as actionbar

yanchenko:
Concerning https://github.com/JakeWharton/ActionBarSherlock/pull/178


JakeWharton:
GitHub notifications were off on this by default for some reason. Never got the initial email... sorry!

Thanks for the updated contribution!

PR #199, opened 2012-01-23
add perm to enable demo installation on Google TV

larham:
the "touch not required" perm does not affect normal handset installation nor change the list of perms shown to users. it enables installation on devices like Google TV that have no touch screen.  (Android default is that touch is required.)


JakeWharton:
Not taking any new things on the `master` branch but I've added this for the next major release. Thanks!

PR #202, opened 2012-01-25
rm classes from android-support-v4 that we override.

mlc:
I have ActionBarSherlock and the support library as libraries of my app using the ant properties library mechanism. The dx step would fail until I made this change.

Thanks!


JakeWharton:
You should be using the compiled .jar of the support-lib plugin in your apps `libs/` folder rather than as a library project.

iNoles:
I can help you to get it to work under apache ant build. 

``` xml
<target name="-pre-compile">
    <unzip src="android-support-v4.jar" dest="${out.classes.absolute.dir}">
        <patternset>
            <exclude name="**/FragmentActivity*,**/FragmentManager*" />
        </patternset>
    </unzip>
</target>
```

This goes in build.xml

JakeWharton:
You should include `**/Fragment.*,**/Fragment$*` too. I seem to have forgotten it in the `pom.xml`...

iNoles:
Heh. it would be like 

``` xml
<exclude name="**/FragmentActivity*,**/FragmentManager*, **/Fragment.*,**/Fragment$*" />
```

?

PR #205, opened 2012-01-26
Fix for totally unqualified activity class names

### LINKED ISSUE #201: android:uiOptions="splitActionBarWhenNarrow" in Activity not working
android:uiOptions="splitActionBarWhenNarrow" gives a split ActionBar when placed in the application node,
but fails to do so when placed in an Activity node.
Switching my app to use native ActionBar support confirms a split ActionBar is shown when placed in the Activity node.

The activity is a FragmentActivity that contains one fragment. This fragment adds the action items to the ActionBar.

I'm using ABS 4.0 Beta 5.

mcnicholls:
Activity class names in the manifest are _supposed_ to be either fully qualified or relative to the current package, but the OS allows totally unqualified names.
This fix is to allow the reading of totally unqualified Activity class names to comply with the OS behaviour.
Closes #201.

### DIFF
--- a/library/src/com/actionbarsherlock/ActionBarSherlock.java
+++ b/library/src/com/actionbarsherlock/ActionBarSherlock.java
@@ -957,7 +957,11 @@ public final class ActionBarSherlock {
                             } else if ("name".equals(attrName)) {
                                 activityPackage = xml.getAttributeValue(i);
                                 //Handle FQCN or relative
-                                if (!activityPackage.startsWith(packageName) && activityPackage.startsWith(".")) {
+                                if (!activityPackage.startsWith(packageName)) {
+                                    //Handle unqualified package name
+                                    if(!activityPackage.startsWith(".")) {
+                                        activityPackage = "." + activityPackage;
+                                    }
                                     activityPackage = packageName + activityPackage;
                                 }
                                 if (!thisPackage.equals(activityPackage)) {
--- a/library/src/com/actionbarsherlock/ActionBarSherlock.java
+++ b/library/src/com/actionbarsherlock/ActionBarSherlock.java
@@ -957,9 +957,9 @@ public final class ActionBarSherlock {
                             } else if ("name".equals(attrName)) {
                                 activityPackage = xml.getAttributeValue(i);
                                 //Handle FQCN or relative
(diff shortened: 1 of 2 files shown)


JakeWharton:
This won't work if the user specified a fully qualified class name in the manifest that is different from their manifest package

E.g.,

``` xml
<manifest
    package="com.jakewharton.whatever">
    <!-- ... -->
    <activity android:name="com.some.library.package.SuperActivity" />
```

Need to add something like`.indexOf('.', 1) == 0` to the conditions.

mcnicholls:
Ahh I hadn't considered that case.

How does it look now?

JakeWharton:
Just thought of another case.

``` xml
<manifest package="com.jakewharton.whatever">
    <!-- ... -->
    <activity android:name=".ui.ThisAndThat" />
```

So the three cases that need to be handled are:
- Fully qualified class name (e.g., `android:name="com.my.package.TheActivity"`)
- Relative class name that starts with '.' (e.g., `android:name=".ui.ThisAndThat"`)
- Unqualified class name (e.g., `android:name="TechnicallyWrongButStillWorksActivity"`)

I think we might be able to get away with this by simply adding the `startsWith('.')` check back in as the first condition (since it's the most common case, hopefully) and then or-ing the `indexOf`.

I'm going to write up some test cases for this one today...

mcnicholls:
Yes of course, I hadn't considered the relative case that has multiple '.''s in it.

I had a look to see if I could add a test case for this fix, but wasn't too sure how to go about it.
Will be interested to see how you add the tests.

JakeWharton:
Implemented this myself [here](https://github.com/JakeWharton/ActionBarSherlock/blob/ea1b0573c21603bfe6ec205d608403174709a8b2/library/src/com/actionbarsherlock/internal/ActionBarSherlockCompat.java#L703-714) along with a juggernaut commit. Tests are [here](https://github.com/JakeWharton/ActionBarSherlock/blob/ea1b0573c21603bfe6ec205d608403174709a8b2/library/test/com/actionbarsherlock/internal/ManifestParsingTest.java) which are run via Maven (and pass!).

Thanks for your help on this.

mcnicholls:
That all looks great.

Will know what I am doing next time if I need to add a test.

PR #215, opened 2012-02-06
Get overflow listviews top/bottom padding, instead of its top/bottom. Should close #181

### LINKED ISSUE #181: Overflow Measurement Intermittently Filling Screen
That pesky @scottkennedy discovered this gem:

SimonVT:


JakeWharton:
Damn it's so obvious now that I look at it! I could literally reach through the internet and hug you right now.

PR #218, opened 2012-02-06
4.0 wip

MarcusWolschon:
Your MenuWrapper returned NULL on Android 4.0.x for the UP MenuItem (itemID==android.R.id.home).
This resulted in onOptionItemSelected being called with NULL.
I left the 2 log-statements in there to show when exactly this happens and also added some code to make this bullet proof in case a future Android version changes it's behavior slightly (e.g. comparison by instance instead of itemID of MenuItems failing).


JakeWharton:
Thanks! I pulled in your changes manually since the pull request was against the wrong branch. I also chose to be a bit less forgiving (at least for now) when looking up items. If I see a lot of errors coming up I'll change it to be more forgiving but I prefer to fail fast for now.

MarcusWolschon:
To fail fast I found it very rewarding to do strict input checking whereever possible.
At least checking for null arguments.

PR #261, opened 2012-02-14
Remove odd null check

kevinsawicki:
Remove unneeded checks on action view being null and simply return it either way

### DIFF
--- a/library/src/com/actionbarsherlock/internal/view/menu/MenuItemImpl.java
+++ b/library/src/com/actionbarsherlock/internal/view/menu/MenuItemImpl.java
@@ -567,10 +567,6 @@ public final class MenuItemImpl implements MenuItem {
     }
 
     public View getActionView() {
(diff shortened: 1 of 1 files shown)


JakeWharton:
Woah! I'm glad you sent this. I need to re-meld `MenuItemImpl` as it looks like it still seems to be from ABS v3.x.

Here's the ICS version of this method: https://github.com/android/platform_frameworks_base/blob/master/core/java/com/android/internal/view/menu/MenuItemImpl.java#L570

JakeWharton:
Wait, nevermind! This is on `master`. All development is now on the `4.0-wip` branch and I'm about 24 hours away from a release candidate for v4.0. Thanks for the pull request but the version 3.x is about to be heavily deprecated in favor of the new version.

kevinsawicki:
Oh okay, sorry about that, keep up the great work!

PR #287, opened 2012-02-18
Ignore classes/ folder

zhen9ao:
Ignore the classes/ folder, which will generate after demos run in IntelliJ IDEA.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ lint.xml
 *.iml
 *.ipr
 *.iws
+classes
 
 #Maven
 target


ianpartridge:
Might be better to just whitelist the stuff we do want managed in git, rather than whack-a-mole-ing forever:

.gitignore:

*
!library/
!README.md

etc..

JakeWharton:
With whitelisting we have to worry about people not being able to figure out how to add stuff. I imagine that even I would forget that I had to do it after a while. Ignoring everything is pretty much the only thing we can do.

zhen9ao:
Maybe I think in the git thery, it's all about blacklisting, but not whitelisting, am I right?

PR #306, opened 2012-02-26
Use the faster FloatMath!

scottkennedy:


hrj:
Are you sure that FloatMath is fast?

Check this discussion: http://stackoverflow.com/questions/15143735/using-floatmath-or-math-and-a-cast

PR #313, opened 2012-02-27
Use TypedArray when parsing menu xml. Closes #293.

### LINKED ISSUE #293: Icon not shown if android:icon references a theme attribute
^

SimonVT:
I only tested this on my own project, but it fixed any issues I was having. Tested on 2.1 and 4.0.

### DIFF
--- a/library/res/values/abs__attrs.xml
+++ b/library/res/values/abs__attrs.xml
@@ -231,4 +231,100 @@
     <declare-styleable name="SherlockActionMenuItemView">
         <attr name="android:minWidth" />
     </declare-styleable>
+
+    <!-- Base attributes that are available to all groups. -->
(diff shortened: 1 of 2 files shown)


MichaelRocks:
After this fix the logcat on 3.X devices is full of warnings like
`08-21 00:45:09.766: W/Resources(19505): Converting to string: TypedValue{t=0x1/d=0x103011c a=2 r=0x103011c}`
`08-21 00:45:10.695: W/ResourceType(19505): Requesting resource 0x103011c failed because it is complex`
`08-21 00:45:10.695: D/dalvikvm(19505):   JDWP invocation returning with exceptObj=0x40675470 (Landroid/content/res/Resources$NotFoundException;)`

I'm not sure there are no wanings on other versions of Android. Also it seems that this fix is the reason of #540.

Issue #318, opened 2012-02-29
Crashes if a sub-menu is added and 'showAsAction' is not set on the su-menu


guidedways:
The following XML causes ActionBarSherlock to crash while parsing showAsAction (as it can't find it):

<menu xmlns:android="http://schemas.android.com/apk/res/android">
    <item
            android:id="@+id/menu_delete"
            android:title="@string/delete"
            android:icon="@drawable/ic_action_delete"
            android:showAsAction="always"
            >
        <menu>

```
        <item android:id="@+id/delete_confirm"
              android:title="@string/delete"
              android:icon="@drawable/ic_action_deletelight" />

        <item android:id="@+id/cancel_delete"
              android:title="@string/cancel"
              android:icon="@drawable/ic_action_cancellight" />
    </menu>
</item>
```

</menu>

I then added 'android:showAsAction="never"' to the sub-menus and the crashing stopped.

JakeWharton:
Dupe #317.

Issue #317, opened 2012-02-29
No android:showAsAction attribute on submenu items crashes MenuInflater


yukuku:
Using post-RC1, commit 17cd662d5d10583cbfc8ec09e446ab45e498dff0 
Doesn't happen on RC1.

My menu XML file is:

``` xml
<?xml version="1.0" encoding="utf-8"?>
<menu xmlns:android="http://schemas.android.com/apk/res/android">
    <item android:id="@+id/menuNavigate" android:icon="@drawable/ic_menu_navigate" android:showAsAction="always" android:title="Navigation Menu" android:orderInCategory="9">
        <menu>
            <item android:id="@+id/menuH" android:title="H" />
            <item android:id="@+id/menuE" android:title="E" />
            <item android:id="@+id/menuM" android:title="M"/>
            <item android:id="@+id/menuA" android:title="A" />
            <item android:id="@+id/menuV" android:title="V" />
            <item android:id="@+id/menuB" android:title="B" />
        </menu>
    </item>
</menu>
```

When I inflate it using getSupportMenuInflater(), I got the below error.

If I add showAsAction to the submenu items, it no longer crashes, for example:

``` xml
<item android:id="@+id/menuH" android:title="H" android:showAsAction="never" />
```

Logcat when error happens:

```
02-29 16:03:46.713: E/AndroidRuntime(14202): java.lang.NumberFormatException: unable to parse '@16973911' as integer
02-29 16:03:46.713: E/AndroidRuntime(14202):    at java.lang.Integer.parse(Integer.java:383)
02-29 16:03:46.713: E/AndroidRuntime(14202):    at java.lang.Integer.parseInt(Integer.java:372)
02-29 16:03:46.713: E/AndroidRuntime(14202):    at com.android.internal.util.XmlUtils.convertValueToInt(XmlUtils.java:122)
02-29 16:03:46.713: E/AndroidRuntime(14202):    at android.content.res.TypedArray.getInt(TypedArray.java:254)
02-29 16:03:46.713: E/AndroidRuntime(14202):    at com.actionbarsherlock.view.MenuInflater$MenuState.readItem(MenuInflater.java:353)
02-29 16:03:46.713: E/AndroidRuntime(14202):    at com.actionbarsherlock.view.MenuInflater.parseMenu(MenuInflater.java:146)
[...]

PR #320, opened 2012-03-01
4.0 wip

Sleepybear:
Fix for @SimonVT's change in menu inflater that was refferrincing an attribute in HC+ only. Not sure why git thought I changed lines to be the exact same thing when I didn't even touch them... but there it is.

If you can think of a more elegant solution let me know, but wanted to at least get 4.0-wip runnable on <3.0 again.

Refs #317,#318


Sleepybear:
Simon's solution is better.

PR #321, opened 2012-03-01
Support for ADT 17 Lint checks

kuglimon:
On the next version of ADT Lint will check for usage of methods from older platforms. This is by default turned to produce an error which of course breaks ABS. Users shouldn't need to care about this.

A thing to note:

The Lint check seems to fail on anonymous classes, it will complain about usage even if the classes are marked with @TargetApi.


kuglimon:
Ah damn went to the wrong branch. Closing.

PR #322, opened 2012-03-01
Support for ADT 17 Lint checks

kuglimon:
On the next version of ADT Lint will check for usage of methods from older platforms. This is by default turned to produce an error which of course breaks ABS. Users shouldn't need to care about this.

A thing to note:

The Lint check seems to fail on anonymous classes, it will complain about usage even if the classes are marked with @TargetApi.

### DIFF
--- a/library/src/com/actionbarsherlock/TargetApi.java
+++ b/library/src/com/actionbarsherlock/TargetApi.java
@@ -1,0 +1,15 @@
+
+package com.actionbarsherlock;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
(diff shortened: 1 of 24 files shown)


JakeWharton:
I replied via email but apparently it was eaten by the lack of cell service crossing Utah and Nevada.

From what I gather you are giving everything API 14 compatibility which is very much not the case and defeats the purpose of these new lint checks. Anything not suffixed with 'Native' or 'Wrapper' is written for API 7 with conditional blocks in some cases for newer APIs.

A library like this needs to leverage this new feature to its fullest extent since it is required to work on so many platforms. I was going to hold off on integrating this stuff until after the new ADT was released. If you'd like you can update your branch (and thus this request) or just hold off and I'll get to this soon.

kuglimon:
I see, a bit of misunderstanding on my part.

I'll look in to it more thoroughly during the weekend.

PR #323, opened 2012-03-01
Check type of showAsAction before reading value

SimonVT:
Fixes issue in GB where it thought showAsAction was a reference to a drawable if no value was set.


JakeWharton:
Trying to manage this crap from my phone while driving through the Rockies... not fun.

SimonVT:
Don't get arrested for driving and merging

mosabua:
Stop it .. enjoy the rockies and switch off your phone.

PR #328, opened 2012-03-04
Minor Typo in onAttach Exception message in SherlockListFragment

passy:


JakeWharton:
Good catch. Went a bit further and corrected them in every fragment base class as well.

PR #332, opened 2012-03-05
Update LICENSE.txt

gentlecat:
Shouldn't your name be in licence instead of template, Jake?

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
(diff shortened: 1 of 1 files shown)


JakeWharton:
Hmmm... maybe.

I left it as is originally since it was part of the official document and instead describes how to annotate your stuff with the license. The license itself isn't really mine. I put my name in there when used in the `README` and other places though.

Plus I don't want to change it since it is the only file that still shows the 'Initial commit' message. :)

gentlecat:
Okay :)

PR #337, opened 2012-03-07
Add new SeriesGuide screenshots, update to Google play links

UweTrottmann:
Changed the screenshots on the samples page, updated all Android Market links on that page to Google play.


JakeWharton:
The website is stored in `website/resources/` in the `dev` branch and gets deployed automatically by maven (so this would be overwritten). I'd **really** appreciate if you'd make the changes there and submit a pull request against `dev` before 11:55PM PST so I can get it in the 4.0 release. I'm not sure I'll have time to do it manually before then.

JakeWharton:
I have no idea what timezone it is where you are. You're probably asleep...

JakeWharton:
Nevermind. I'm changing the samples page a bit until the full website rewrite.

UweTrottmann:
Got up early, is it too late yet? I'm Central European Time (Germany). My clock says it's close to eleven on PST.

JakeWharton:
I dropped screenshots for now from the samples page as they take far too long to get working properly and are far too manual of a process. They'll be part of an upcoming full-site rewrite with a mini-market place of ABS apps (hopefully).

JakeWharton:
You're still front and center (and bottom...) on the index.html though!

UweTrottmann:
Awesome :D I just saw the old screenshots and had to do something about that... The mini-market idea sounds great.
Maybe every app could also add a 'tag' in their description on 'Google Play' so you could search for ActionBarSherlock apps.

JakeWharton:
That's not a bad idea either. I'll have to experiment.

It's too bad the Market doesn't have a proper API for fetching screenshots and leading images and the like. Would make the creation of something like that super easy.

PR #344, opened 2012-03-08
Update some old links to Android Market

gentlecat:


JakeWharton:
Good eye! Thanks for updating the request. We'll have a .1 release soon enough that this will go out with.

PR #346, opened 2012-03-08
Replaced instance of Android Market with Play Market

veeti:
You forgot one instance of "Android Market" on the front page :p


JakeWharton:
This has already been addressed in the `dev` branch. Thanks for the pull request, though.

nelenkov:
It seems it's actually called the 'Play Store', although 'store' doesn't show up on https://play.google.com/store. The Android launcher shortcut is called 'Play Store' though. Go figure.

PR #350, opened 2012-03-09
Overflow menu doesn't get disabled in action mode

veeti:
The overflow menu (if any) can still be activated on 2.3 and below while in an action mode, unlike the native behavior. I've got a sample up [here](https://github.com/veeti/ABSActionModeInconsistency) and a 2-line fix that I _think_ works properly on this pull request.

Thanks for the great library!

### DIFF
--- a/library/src/com/actionbarsherlock/internal/ActionBarSherlockCompat.java
+++ b/library/src/com/actionbarsherlock/internal/ActionBarSherlockCompat.java
@@ -318,6 +318,9 @@ public class ActionBarSherlockCompat extends ActionBarSherlock implements MenuBu
     public boolean dispatchPrepareOptionsMenu(android.view.Menu menu) {
         if (DEBUG) Log.d(TAG, "[dispatchPrepareOptionsMenu] android.view.Menu: " + menu);
 
+        if (mActionMode != null)
+            return false;
+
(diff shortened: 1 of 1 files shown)


JakeWharton:
You missed a couple spaces :)

veeti:
Oops: I'm still a bit new to IntelliJ IDEA (although it has made development a million times less painful already). Unless both that and Notepad++ are lying to me, it should be right now.

JakeWharton:
You're saying that this can be activated using the menu key, correct? Not by some other means.

veeti:
Yes.

JakeWharton:
Hmm this was supposed to be handled by [this code](https://github.com/JakeWharton/ActionBarSherlock/blob/master/library/src/com/actionbarsherlock/internal/ActionBarSherlockCompat.java#L420-426). I'll have to whip up a sample locally and see what's going on...

JakeWharton:
I cannot reproduce this. Please paste the code for a demo activity which exhibits this behavior.

veeti:
[Here](https://github.com/veeti/ABSActionModeInconsistency).

JakeWharton:
Whoops. I was only testing the forced-overflow. Thanks!

PR #357, opened 2012-03-09
Cleans up some warnings

mrenouf:
These all seem like redundant and unnecessary casts. Can just remove?


JakeWharton:
None of this was my code. Woot!

Merged to `dev`.

PR #360, opened 2012-03-10
reduce number of exceptions thrown

bvgastel:
during memory analysis of my app, saw that lots of exceptions were thrown in this piece of code due to OutOfBoundsException, so looked at it and reworking it. not sure if it is correct though.

It is my first pull request, so if you want more info or additional changes or something, let me know.


JakeWharton:
On what version of Android are you seeing the exceptions?

bvgastel:
2.3.7. But as the exceptions are caught and handled, they only show up in a memory profiler and not in the console. While scrolling through a list view, most allocations are exceptions from that method.

hrk:
diff looks fine to me.

JakeWharton:
I'm just curious about the usage that's causing an exception. That code is copied directly from the Android source and I'd need a strong case in order to change it.

Can you provide a small demo app or something that exhibits the behavior you are seeing?

hrk:
I don't want to bust into this, but the source of the exception is obvious just looking at the call to "getChildAt(index -1)".
When the index is 0, it will be called with an argument of -1. Going up the chain, the implementation is in ViewGroup#getChildAt() and looks like this: "try accessing an array at the given index, returning null if it goes beyond the bounds of the array".

http://grepcode.com/file/repository.grepcode.com/java/ext/com.google.android/android/1.5_r4/android/view/ViewGroup.java#ViewGroup.getChildAt%28int%29

JakeWharton:
I get why the exception is thrown, I was just wondering how it was being triggered repeatedly enough that it would cause an issue. I have to assume his entire window is being invalidated a high rate which would cause this code to be triggered.

I'll submit a more efficient method to AOSP and once it's reviewed and accepted by Google I will update the library.

hrk:
I'm sorry if I sounded like a teacher, it was not my intention. :)

bvgastel:
I get the problems in a piece of code with (temporary) lots of invalidates, as an incoming event is displayed. The size of the incoming event grows, so triggers an invalidate. Still want the sample code or do you know enough?

I looked up the code of API 15, and indeed saw the same code. It is strange though that hasDividerBeforeChildAt() in the superclass of ActionMenuView, namely IcsLinearLayout, handles the border cases properly. (or LinearLayout in AOSP).

bvgastel:
the memory allocations of the exceptions slows down the animation in my app. it there any progress with the update to ASOP?

iNoles:
Do you use Eclipse Memory Analyzer?

under Listview, Android always recycling any child view in every time you are scrolling.

JakeWharton:
@bvgastel I'll try to get to it within the next week or so. Not sure how long it'll take to get approved and accepted but hopefully it can make it into a 4.1 release for ABS. You'll still see the problem on ICS and up even after it's fixed, however.

bvgastel:
@iNoles yeah, in conjunction with the DDMS memory allocation tracker. I know listview reuses the views, but the problem is that when a view changes in size, an IndexOutOfBounds exception is thrown (and handled) deep inside the code. This exception can easily be avoided by using a simple bound check.

hrk:
@JakeWharton I don't understand why the problem will be present on ICS if @bvgastel applies his patch (the bound check) to ABS, even if in his own forked repository (It's what I did with mine, by the way). Having said that, merging the bound check in the library seems reasonable to me, knowing that there are broken releases of Android which could possibly never get an update. I can't see why it shouldn't be merged. Please, don't take this as an offence or disrespect, just wondering due to my limited knowledge of the library and your approach.

JakeWharton:
Got AOSP contributions up and going again tonight (see 30c1ded12bfbbf2d7cc9708bbee9f1961cd81fdf) so I'll try to get to this by the weekend. Need to review the code more closely before submitting to AOSP.

JakeWharton:
@hrkdroid Because ICS uses the native action bar and thus will always have this problem until ABS 5.0 where we steal the sources from Jelly Bean and backport it to work on Android 2.x - 4.x and the native only on Android 5.0+.

hrk:
@JakeWharton maybe I see the light... is library/src/com/actionbarsherlock/internal/view/menu/ActionMenuView.java one of the files only used in <4.0?

JakeWharton:
Besides the menu wrappers, the only class in `com.actionbarsherlock.internal.*` that's used on 4.0+ is `ActionBarSherlockNative.java`.

hrk:
Thanks for the clarification!

jfcogato:
I have the same problems up there. Jake, you've going to upload a new version with this bug cleared? you are making a really good works here ;)

JakeWharton:
https://android-review.googlesource.com/39998. Applied from #511.

PR #371, opened 2012-03-12
Fix menu item wrapper

wuman:
Removing the @Override annotation will allow the MenuItemWrapper class to be linked successfully on devices earlier than level 14.


wuman:
@JakeWharton Your patch aa9ec27 won't work because you're trying to instantiate the native OnActionExpandListener at MenuItemWrapper construction time.  Doing a lazy instantiation inside the setOnActionExpandListener() method will eliminate the problem.

Please reopen this so that I can send another patch.

PR #384, opened 2012-03-16
Supporting tabs with action modes on pre-ICS phones

yalon:
Alpha property setter/getter is missing from ScrollingTabContainerView. Added delegate methods to mTabLayout to solve this.


JakeWharton:
Thanks! I chose to implement a bit differently to reduce the changes to code that is directly from Android to make future merges more straightforward.

Issue #394, opened 2012-03-22
Update to r7 Support Library


JakeWharton:
Lets just spearhead this issue's creation...

https://plus.google.com/108284392618554783657/posts/XN5qqHPwnRQ

PR #395, opened 2012-03-22
Adapted Action Bar Sherlock to the compat library changes

jleidgens:
A new compatibilty jar is available and since the new build configurations now compares jar files I added the new jar to the Actionbar Sherlock Project. 

In the new compatiblity jar the visibility of one off the menu methods changed. I also fixed this error.


JakeWharton:
I really do appreciate the effort but the main development and compilation of the library is done using Maven and there is no r7 artifact yet. Once Google updates AOSP with the new changes I can start the process of getting it up into central and then (and only then) will I be able to update the `dev` branch to support it for the next release.

See the link in #394 if you need a solution before then. It basically describes exactly what you've done, though.

PR #410, opened 2012-03-28
Updated Android Market link to point to the new Google Play URL

casidiablo:
I guess Google will keep the redirection from the old site to the new one, but I think it is better to have the new one.


JakeWharton:
Thanks! Merged to `dev` for next release.

PR #421, opened 2012-03-31
Update android-support-v4.jar to revision 7

beshkenadze:
Fix visibility from protected to public SherlockFragmentActivity.supportInvalidateOptionsMenu


JakeWharton:
Waiting on Google. See #394.

PR #423, opened 2012-04-01
Removed unnecessary Runnable field

casidiablo:
`mTabSelector` field (`Runnable`) is never assigned.


JakeWharton:
I would prefer if you commented them out with `//UNUSED` so that in future merges with AOSP updates it's obvious that they were removed deliberately and are not re-added.

JakeWharton:
Thanks!

PR #428, opened 2012-04-04
IcsLinearLayout lacks a getter/setter for padding

nacitar:
I intend on using IcsLinearLayout in my application, and thought I'd add these methods while I was at it; would prefer it to be in your sources.


JakeWharton:
Be careful using anything in the `internal` package as there is no guarantee its API will not change dramatically or disappear completely.

Merged to `dev`.

PR #441, opened 2012-04-18
Declare support-library as provided in ActionBarSherlock-library

xperimental:
Would it make sense to mark the support-library as provided in the ActionBarSherlock library pom? That way every application could decide which support-library to use.

In my case that would be a library which uses MapActivity as super class for FragmentActivity, because I want to display fragments and a Google-Maps map in an activity.

I've already seen, that your ViewPagerIndicator library (which is also used in that application) also declares the support-library as provided.

### DIFF
--- a/library/pom.xml
+++ b/library/pom.xml
@@ -24,6 +24,7 @@
 		<dependency>
 			<groupId>com.google.android</groupId>
 			<artifactId>support-v4</artifactId>
+			<scope>provided</scope>
 		</dependency>
 
 		<dependency>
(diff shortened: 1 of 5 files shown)


JakeWharton:
I would much rather leave it as a hard dependency. You can always use an exclusion to avoid it.

``` xml
<dependency>
    <groupId>com.actionbarsherlock</groupId>
    <artifactId>library</artifactId>
    <version>4.0.2</version>
    <type>apklib</type>
    <exclusions>
        <exclusion>
        </exclusion>
    </exclusions>
</dependency>
```

Maven will not include duplicate transitive dependencies and if you have another library which is dependent on a different version you can use the above technique to remove one of them.

xperimental:
Thanks for the hint, that solved my problem.

PR #484, opened 2012-05-19
Make it easier to use in Eclipse

thasmin:
These commits make it easier to use in Eclipse. The first stops Eclipse from saying the version of android-support-v4.jar is wrong. The second puts the project files into the got repo so we can use the library by doing a file -> import -> existing project.


JakeWharton:
Ok, so:
1. I don't use Eclipse. I construct the directory structure in a manner that's consistent what with Android applications and library projects use so that they can be used with both Eclipse and Ant. Any problems in importing are related to the tooling and not the library. I do not believe that including IDE specific files in the repository is correct and it should never be necessary.
2. You cannot upgrade the version of the support library without upgrading the version that Maven uses. I use Maven as a build system because it's an actual build system. Ant is Bash shoved into XML and I refuse to use it. However, the project is perfectly compatible with Ant and you can run `android update project -p .` to get a working Ant-based build. The lack of dependency versioning is not the libraries fault but Ant and ADTs.
3. You are deleting files from `.gitignore` that are fundamental to the compilation in IDEs other than Eclipse which will ultimately result in git attempting to track things that should not be revision controlled. I do not force my IDE on others nor my build system which is why the project works with Ant, Maven, Eclipse, and IntelliJ IDEA and the contents of this file are crucial to maintaining that.

If you require the things which you have contributed in this pull request I suggest you maintain your own fork on GitHub and merge changes from my repository whenever there are updates. I appreciate your contribution but unfortunately these things do not belong in this repository.

thasmin:
Let me explain my use case. I created an open source podcast player called Podax. It uses your libraries because they're extremely good and they make Podax a better app. Without these files, I need to create detailed instructions on how to download the code, set it up to compile, configure Podax to use it as a library, and then get around Eclipse's oddities in order to share my source code. I'm just trying to make it easy to contribute so the world can have a better podcast player on their phone. Eclipse is the officially supported way to create Android apps, so I chose that.

I'd argue that this library doesn't support Ant or Eclipse out of the box. It's easier to get Ant working than Eclipse, but both need extra steps to get the code to compile. The files removed from .gitignore are not preventing people from using other compilers. The version of the Android Support library that I committed is the most recent version. If you are only supporting Maven then just say so. That's fine.

I don't blame the libraries for problems in the tooling. It would be nice if libraries could be self contained on Android and you could distribute APKs. You've clearly put a lot of work in to make some awesome libraries. I'm just asking that there is official support for the officially supported build methods.

Anyway, thanks for making me look up that Maven can do incremental compiles. I'd love to use vim instead of Eclipse but the long compile times were holding me back.

marczych:
You should take Jake's advice: fork this repository and include all necessary Eclipse files so anybody who wants to use it can easily get it set up. You can of course pull in changes whenever you want (which is what you'd have to do without a fork anyway).

I'm also very interested in ditching Eclipse for many reasons. Is it really as easy as:

`android create project ...`
< edit project >
`ant debug`

to do Android development without Eclipse? If you could point me in the direction of some resources that would be great =).

thasmin:
Yeah, that's pretty much it. But Ant compilation takes a long time.

JakeWharton:
The one thing I thought about including was a `build.xml` file for Ant compilation but it still requires you to run `android update project -p .` anyways. Another reason I haven't made any changes is that Tor and Xavier, the tools developers at Google, have been pretty adamant that bundling library projects into .jars is coming. It sadly just keeps getting pushed back. Once that happens you should theoretically be able to distribute ABS as just another .jar in your `libs/` folder and all of these points will be mostly moot.

@marczych http://www.jetbrains.com/idea/features/google_android.html

marczych:
I'd rather not use an IDE at all - I much prefer doing things on the command line. I've been using eclim for a while but of course it still requires eclipse. I'll look more into ant when I get back into Android dev.

Issue #351, opened 2012-03-09
ViewPager Action Bar Items not showing


ericklee:
The ViewPager is set up like this:
http://developer.android.com/reference/android/support/v4/view/ViewPager.html

I have 4 Fragments under the ViewPager and each Fragment may have difference menu item which setShowAsAction is set to SHOW_AS_ACTION_ALWAYS

setHasOptionMenu(true) has put in onCreate method in the Fragment.

The Action Items are working fine when the activity first created. 
But if I flip the fragment quickly, at some point all Action Items will be disappeared.
The Action Items can only be restored if you change the orientation of the phone.

ericklee:
Problem seems solved if I replace 
mActionBar.setSelectedNavigationItem(position) 
with 
mActionBar.getTabAt(arg0).select()
inside method onPageSelected()

mario20340:
I have the same issue and your fix doesnt seem to help me at all... 

I'm using viewpager to switch between 4 actionbar tabs... there is a fragment in every tab with different action menu items defined in onCreateOptionsMenu()... When I switch too fast between tabs (using slide guesture) action menu items suddenly disappear... This is happening on Froyo device... I cannot check on other devices because I dont have any other devices... In emulator I cannont slide that fast to recreate this behaviour.... any help why is this happening

ericklee:
Issue seems still occur.

JakeWharton:
Probably a dupe/related to #331 but I'll leave it open for now.

mario20340:
I noticed that, when I've disabled action bar tabs everything works. Im using your viewpager indicator now insted of actionbar tabs and everything is fine... It seems that is something wrong with actionbar tab and its adapter

ericklee:
Does the latest 4.0.2 version fix this?
I cannot recreate the bug now.

JakeWharton:
It's possible but the fix I put in place for menus wasn't expected to fix this. I haven't been able to reliably reproduce anyways so I remain unsure.

JakeWharton:
I'm fairly certain this is a native bug that deals with mismatch of the asynchronous nature of the `FragmentManager` and the synchronous nature of `invalidateOptionsMenu()`. Will convert `FragmentPagerAdapter` to confirm.

snodnipper:
[...]

Issue #272, opened 2012-02-15
FragmentActivity - options menu not respecting Fragment lifecycle


jgilfelt:
When using the support lib plugin FragmentActivity with a Fragment based ViewPager, I get weird behavior when the fragments provide their options menu to the hosting Activity's action bar:
- The initial menu actions for the 1st fragment loaded into the ViewPager work fine.
- As soon as a 2nd fragment is paged into the UI, the menu actions become unresponsive.
- As soon as a 3rd fragment is paged into the UI, the actions that I tried to press for fragment #2 will fire, and so on.
- Orientation change yields similar results to fragment paging.
- There is a subtle visible artifact when a 2nd fragment is loaded, as if the new action items are being drawn over previous ones.

I am using a standard FragmentPagerAdapter pattern in my code. I have created a simple example project demonstrating this issue from the Google SupportDemos sample. It works flawlessly on my Galaxy Nexus with the native action bar, on anything else I get have the issues described above. I am using the ABS 4 RC1 library & support lib.

http://dl.dropbox.com/u/3982805/SherlockFragmentActivityMenuDemo.zip

JakeWharton:
I'll have to go through the (crazy) callbacks of menu invalidation once more since there is obviously something out of place. Thanks for the sample. I'll try to get it done ASAP.

JakeWharton:
Ok. After far too much debugging of the invalidation callback lifecycle I've found that this bug is in a totally separate area.

In order to support dividers between tabs and action items I backported ICS's version of `LinearLayout` which now supports dividers between children natively. Due to not being in the `android.widget` package and not being able to access methods marked with `@hide`, some things were omitted and a lot was stripped so that the only added functionality was the dividers. It would seem that these two factors has led to a few bugs in layout and touch interaction which cause both this and the touch problem you reported earlier.

I'm going to try to adapt the `IcsLinearLayout` class to extend from `LinearLayout` (rather than just `ViewGroup`) and have its implementation only control drawing the dividers.

If you want a stop-gap solution replace `IcsLinearLayout` with `LinearLayout` in `ActionMenuView.java` and comment out anything about dividers. Or perhaps a brutal (semi-untested) hack might work:

``` java
@Override
public void invalidateOptionsMenu() {
    super.invalidateOptionsMenu();
    getWindow().getDecorView().requestLayout();
}
```

jgilfelt:
Thanks Jake! 

[...]

PR #485, opened 2012-05-20
added known-bugs test case for #272

jgilfelt:
I think there might be regression since ABS 4.0.1 here as the same symptoms in the original issue report which were fixed prior to 4.0.0 can be reproduced in 4.1.0. It may be related to #351 but this behavior is pre-ICS only. In any case I have added an example to the known-bugs project.


JakeWharton:
Thank you! And good job picking up on all the conventions for naming and in the manifest and such.

PR #490, opened 2012-05-22
Cleanup

ChristianKatzmann:
Removed many Java Compiler warnings like missing @override annotations and unnecessary casts

### DIFF
--- a/library/src/com/actionbarsherlock/app/SherlockActivity.java
+++ b/library/src/com/actionbarsherlock/app/SherlockActivity.java
@@ -125,6 +125,7 @@ public abstract class SherlockActivity extends Activity implements OnCreatePanel
         return getSherlock().getMenuInflater();
     }
 
+    @Override
     public void invalidateOptionsMenu() {
         getSherlock().dispatchInvalidateOptionsMenu();
     }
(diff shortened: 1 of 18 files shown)


zhen9ao:
Seems no need of this `cleanup`.

JakeWharton:
What IDE and/or build system are you using where these are warnings? The `@Override` annotation is a compile-time enforcement only and has no effect on the resulting bytecode which is why I'm hesitant to include them. Plus a lot of the code you have added them to is copied directly from the Android sources and would create additional merge conflicts moving forward with future versions of the platform.

ChristianKatzmann:
I'm using Eclipse and missing `@Override` annotations are marked as warnings (at least in my compiler configuration).

I just was confused that some `@Override` annotations were available and some were missing in the ABS codebase, so I pulled this changes.

iNoles:
I think you have 1.5. You could set 1.6

ChristianKatzmann:
That's not the point. Only difference between 1.5 and 1.6 is the fact, that overridden Interface methods should be declared too in 1.6.

JakeWharton:
Needs rebase. Code is about to drastically change with Jelly Bean source import as well. You should consider submitting these upstream rather than to ABS.

PR #492, opened 2012-05-24
Added localized resources

androidmoney:
Added localized resources for the top 10 most common languages for a total of 100 string resources.

### DIFF
--- a/library/res/values-de/abs__strings.xml
+++ b/library/res/values-de/abs__strings.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- 
+/* //device/apps/common/assets/res/any/strings.xml
+**
+** Copyright 2006, The Android Open Source Project
+**
+** Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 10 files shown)


dancefire:
Why this commit is not pulled? It's very handy for apps with multiple language, otherwise {"abs__xxxx" is not translated in xx} error will be thrown during export apk. It doesn't hurt anything, so please pull.

SimonVT:
This should either be a complete commit which includes all languages in AOSP, or people should manually include strings for languages they support. Just adding 10 is kind of half-assed.

Prototik:
Mmm... For HoloEverywhere i develop "Translator" - grabber of strings from AOSP.
May be run once it for ABS?

JakeWharton:
I don't want this in the core library. ABS is large enough. If people truly want this it can exist in a parallel library project.

JakeWharton:
Please re-submit as another top-level module called `library-i18n` with all AOSP languages included. A script to parse used strings and keep this new module up-to-date given a path to an AOSP clone would also be nice.

Issue #495, opened 2012-05-30
dev branch compiles unsuccessful


zhen9ao:
When compile with `mvn -X clean install`, error occur, it says:

> [INFO]
> [INFO] ------------------------------------------------------------------------
> [INFO] Reactor Summary:
> [INFO]
> [INFO] ActionBarSherlock (Parent) ........................ SUCCESS [0.425s]
> [INFO] ActionBarSherlock ................................. SUCCESS [19.251s]
> [INFO] Sample (Parent) ................................... SUCCESS [0.120s]
> [INFO] Sample: Demos ..................................... SUCCESS [33.900s]
> [INFO] Sample: Fragments ................................. SUCCESS [31.126s]
> [INFO] Sample: Known Bugs ................................ FAILURE [23.995s]
> [INFO] Sample: RoboGuice ................................. SKIPPED
> [INFO] Sample: Styled .................................... SKIPPED
> [INFO] ------------------------------------------------------------------------
> [INFO] BUILD FAILURE
> [INFO] ------------------------------------------------------------------------
> [INFO] Total time: 1:49.443s
> [INFO] Finished at: Wed May 30 15:34:31 CST 2012
> [INFO] Final Memory: 19M/84M
> [INFO] ------------------------------------------------------------------------
> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:
> 2.9.1:checkstyle (default) on project sample-known-bugs: An error has occurred i
> n Checkstyle report generation. Failed during checkstyle execution: There are 8
> checkstyle errors. -> [Help 1]
> org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal o
> rg.apache.maven.plugins:maven-checkstyle-plugin:2.9.1:checkstyle (default) on pr
> oject sample-known-bugs: An error has occurred in Checkstyle report generation.
>         at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor
> .java:217)
>         at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor
> .java:153)
>         at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor
> .java:145)
>         at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProje
> ct(LifecycleModuleBuilder.java:84)
>         at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProje
[...]

PR #496, opened 2012-05-30
Fix issue #495

zhen9ao:
Intellij is so helpful with this kind of error, just run `Reformat Code` on the incorrect code.


JakeWharton:
Thanks!

PR #511, opened 2012-06-07
Return early when current index is 0

kevinsawicki:
On Gingerbread and possibly older versions of Android
ViewGroup.getChildAt would catch an exception when a
invalid index was requested.

Adding a guard at this layer prevents many exceptions
from being allocated and thrown away at the ViewGroup
layer.

https://github.com/android/platform_frameworks_base/blob/android-2.2_r1/core/java/android/view/ViewGroup.java#L3057


JakeWharton:
I like this, but I'll have to get it reviewed and accepted to AOSP before I can merge here since it's code that is from Android. This also relates closely to #360. I have a boatload of free time next week so I'll for sure do it then.

kevinsawicki:
Yeah, apologies, I didn't see that pull request you mentioned which is a similar fix to the same issue.

JakeWharton:
https://android-review.googlesource.com/39998

Issue #240, opened 2012-02-09
setSelectedNavigationItem Does Not Work in Collapsed Tab Navigation


mykola-dev:
Another old issue. I use ViewPager with tabs (code based on Support4Demo). When I flip page, I use this code in PagerAdapter to update tab selection:

``` java
public void onPageSelected(int position) {
 mActionBar.setSelectedNavigationItem(position);
}
```

This works great until I add few actionbar items and tabs collapsed to expandable list (it happens in landscape mode). Then selection of current tab just didnt work. Is I should use some other method instead of setSelectedNavigationItem()?
This issue appears in beta5 and beta6

JakeWharton:
This is a platform bug. [This code](https://github.com/android/platform_frameworks_base/blob/master/core/java/com/android/internal/widget/ScrollingTabContainerView.java#L168-179) doesn't not check if `mTabSpinner` is being used. It always assumes `mTabLayout`.

Submitted a patch to AOSP: https://android-review.googlesource.com/32492

mykola-dev:
I see. Thanks Jake.
Will try to reduce my tabs amount and width :)

toulouse:
@deviant-studio Here's a workaround that may work for you: https://gist.github.com/2657485

commonsguy:
FWIW, since there didn't seem to be an official issue filed for the underlying Android bug, [I went ahead and filed one](http://code.google.com/p/android/issues/detail?id=38500).

mohitum007:
I was also experiencing the same issue. Making a lot of research I found the solution here: 
    https://gist.github.com/toulouse/2657485
I hope this may help someone else too.

Mohit Sharma
http://mohitum.wordpress.com

PR #512, opened 2012-06-07
Fix dropdown selection when navigating to tab using getSupportActionBar().selectTab(tab);

abuchanan-grio:
This fixes a problem with tabbed navigation (ActionBar.NAVIGATION_MODE_TABS) when the tabs have been collapsed into a dropdown menu.  Currently, programmatically navigating to a tab using getSupportActionBar().selectTab(tab) does not correctly set the selected item on the dropdown.  The previous tab title remains displayed, and the UI is left in an inconsistent state where the actual tab does not correspond to the tab selected in the dropdown.

### DIFF
--- a/library/src/com/actionbarsherlock/internal/widget/ScrollingTabContainerView.java
+++ b/library/src/com/actionbarsherlock/internal/widget/ScrollingTabContainerView.java
@@ -178,6 +178,8 @@ public class ScrollingTabContainerView extends NineHorizontalScrollView
                 animateToTab(position);
             }
         }
+        if(mTabSpinner != null)
+            mTabSpinner.setSelection(position);
     }
 
     public void setContentHeight(int contentHeight) {


JakeWharton:
This bug is present in the native action bar so you'd only be fixing it for when the compatibility implementation is used. Since creating different behavior across Android versions is much worse than remaining consistent I cannot patch it yet. See #240.

abuchanan-grio:
On the [Android change](https://android-review.googlesource.com/#/c/32492/1) it sounds like they're waiting on feedback from you ("Jake, do you have any comments to provide based on Adam's review? Thanks!") based on [the inline comments that Adam made](https://android-review.googlesource.com/#/c/32492/1/core/java/com/android/internal/widget/ScrollingTabContainerView.java):

> This will result in a requestLayout and setTabSelected is called as part of onMeasure in this view. Requesting layout during a layout pass can cause problems.
> We should probably do something something more direct instead of setTabSelected to animate to the currently selected tab after the size has changed to avoid this.

Any news on whether a different approach will work better?

Issue #523, opened 2012-06-25
Concatenating String in loop


bepcyc:
Package package com.actionbarsherlock.internal.nineoldandroids.animation contains 3 classes (KeyframeSet, ObjectAnimator and ValueAnimator) which are concatenating a String variable in a for loop inside toString() method. 

That's a bad practice, you need to use StringBuilder instead. 

I guess it's a product of some code generation.

JakeWharton:
Is `toString` ever called for any of these? They are likely there for debug purposes where slowness and String concatenation aren't really a concern.

bepcyc:
I don't see direct usages, but other toString() methods are already implemented with StringBuilder.
Anyway, I already made a patch ;)

JakeWharton:
Explanation on #523.

PR #524, opened 2012-06-25
Closing Issue #523.

bepcyc:
Closing Issue #523.


hasanakyurek:
android 4.0.4 getSupportedActionBar() return null. how to fix this problem?

SeanPONeil:
getSupportActionBar()**

JakeWharton:
This code is copied from NineOldAndroids which is copied directly from Android. I prefer not modifying copied code unless it's absolutely necessary because it means more changed lines to merge when new versions of Android come out. As `toString` would only be called by someone debugging I don't think it's worth making the change.

PR #528, opened 2012-06-27
android-support-v4.jar removed from apklib

olekp:
android-support-v4.jar from libs folder makes problems in Idea IDE. It's unnecessary for projects using maven, so can be safely removed from apklib package


DHuckaby:
Removing the jar would make it a real pain for new Eclipse users.

JakeWharton:
It only removes it from the apklib for Maven users. Nice find. I'll get this merged into `dev` today.

JakeWharton:
Merged to `dev`. Thanks!

PR #530, opened 2012-06-27
FragmentTransaction is null for non-sherlock activities

Vandalko:
Sherlock is passing FragmentTransaction instance in tab callbacks now(since v4). But there is strange instance class check here [ActionBarImpl.java:509](https://github.com/JakeWharton/ActionBarSherlock/blob/master/library/src/com/actionbarsherlock/internal/app/ActionBarImpl.java#L509) and here [ActionBarWrapper.java:322](https://github.com/JakeWharton/ActionBarSherlock/blob/master/library/src/com/actionbarsherlock/internal/app/ActionBarWrapper.java#L322).

There is no point to do that check because you can just use FragmentActivity and obtain FragmentManager:

``` java
        if (mActivity instanceof FragmentActivity) {
            trans = ((FragmentActivity)mActivity).getSupportFragmentManager().beginTransaction()
                    .disallowAddToBackStack();
        }
```

### DIFF
--- a/library/src/com/actionbarsherlock/internal/app/ActionBarImpl.java
+++ b/library/src/com/actionbarsherlock/internal/app/ActionBarImpl.java
@@ -37,7 +37,6 @@ import android.view.accessibility.AccessibilityEvent;
 import android.widget.SpinnerAdapter;
 import com.actionbarsherlock.R;
 import com.actionbarsherlock.app.ActionBar;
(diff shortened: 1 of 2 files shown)


JakeWharton:
Patch has spacing problems which fail the build. Please fix.

JakeWharton:
Merged to `dev` branch. Will release soon. Great find, thanks!

PR #548, opened 2012-07-12
Update support library to revision 9.

pocmo:
Non-Maven users will automatically import the support library from the libs folder. The build tools will complain if this version is not the same version as referenced by other library projects.

The best compromise may be to use the latest version (r9) but then ActionBarSherlock would ship with different support library versions when using maven or not. For some reason r7 is the latest version in the central Maven repository. Is there any reason for ActionBarSherlock to use r7 besides Maven? Smoke test looks good. :)


JakeWharton:
It will be updated very soon. The r9 sources just got released in AOSP which means I can deploy a new Maven artifact.

I couldn't accept this anyways since you didn't update any of the sample projects dependencies.

Issue #365, opened 2012-03-11
Custom tab view vanishes after orientation change


shaversport:
After setting a custom view on an actionbar tab via setCustomView,  and using 

android:configChanges="orientation|keyboard|keyboardHidden"

and changing screen orientation, the custom icon disappears and will not reappear until the application is completely shutdown and restarted.

shaversport:
This is for abs 4.0

Program that recreate the issue can be found here: http://bit.ly/wIsVLy

JakeWharton:
Closing as part of #279. I **strongly** recommend you avoid handing configuration changes as [per the documentation suggests](http://developer.android.com/guide/topics/manifest/activity-element.html#config). I have some ideas on how to fix this but it's not top priority at the moment.

shaversport:
Hey Jake, Thanks for the recommendation.

I've taken your advice and removed the configChanges but I find that the custom view still disappears after an orientation change.  Since this happens in ICS as well, this is not an ABS problem.  (Differently than before, the view will reappear if I rotate back.) Can you think of any reason the custom view would not be shown?  Is this an android bug?

Thanks.

shaversport:
Just upgraded to 4.1 and I don't have this issue with abs or 4.0 anymore.  Does that make any sense at all?

JakeWharton:
Nope. But I'll allow it!

shaversport:
No, it doesn't make sense!

Sorry, last  comment was incorrect.  I forgot to first move to the tab with the custom view before doing the orientation.  Doing that shows the problem still exists in abs and in ICS.

JakeWharton:
Why must you toy with my emotions like this! I'll give it another look with config change handling off and see what I can come up with.

marianeum:
Actually, this also happens when the actionbar is not in stacked mode, the search view gets expanded and the tabs no longer fit.
the selected custom tab view disappears, sometimes the tabs change to a list navigation (which is fine) and when i try to change the fragment from there both items disappear and never come back until the activity is being recreated

ericps:
I experience this with ActionBarCompat and customviews, the selected tab disappears on rotation.

seem to have solved it for myself by removing the configChanges from my activity android:configChanges="keyboardHidden|orientation|screenSize"

https://code.google.com/p/android/issues/detail?id=43213

pamepros:
@ericps yes, that's what's breaking the tab, but what if I still want it to rotate ? I still need the line:
[...]

Issue #279, opened 2012-02-16
Actionbar changes are not properly handled if activity has android:configChanges="orientation" 


rciovati:
I have one activity that has 

``` xml
android:configChanges="orientation"
```

and in the menu has 

``` xml
android:showAsAction="ifRoom|withText"
```

The expected behaviour is that when the screen is in portrait i see just the icon and when it is in landscape i see also the text at the right of the action bar icon.
It works on 4.0.3 but Android 2.3.7 when i rotate the screen it doesn't update the action bar icon.

JakeWharton:
I think I _may_ have fixed this already this morning (but forgot to push). Would you mind grabbing the latest `library/` from `4.0-wip` and giving it a try?

JakeWharton:
Here's the SHA of the possible fix: 656419d648373c98d94a4bb5b7ccf294f628ccce

**edit:** On second thought, this would only be for pre-2.2. I'll look into it...

rciovati:
It's pretty strange. I just tried it on:

-2.2 emulator: ok
-2.3.3 emulator: ok
-2.3.7 device: not woking

I will take a look more in detail.

JakeWharton:
Alright there's definitely something wonky going on. I've never dealt with handling config changes before but the emulator behaves differently from real devices. Perhaps it causes a keyboard change on rotation as well or something.

In any case. Now that I've temporarily removed a broken `IcsLinearLayout` class for the action item display I can see things more clearly. Everything is definitely behaving properly when the callbacks are firing. I just have to determine whether or not they fire as expected.

The documentation for `View` says that on API 8 and above the entire view hierarchy will receive configuration change callbacks and on anything less than 8 I dispatch them myself through the tree. I'll have to flash 2.3, 2.2 and 2.1 on my Nexus One to try (or figure out how to get the emulator to not recreate on rotation).

If you want, you can grab the `library/` from the latest in the `4.0-wip` branch and try again to see what you get. Just be careful because anything with animation (e.g., action modes) will force close your app at the moment...

JakeWharton:
Got it working on the emulator. Needed `keyboardHidden` apparently...

rciovati:
I noticed that the emulator re-create the activity when i rotate the screen, also if i have

``` xml
android:configChanges="orientation"
```

so it's normal that it works.

raymanzhang:
Don't know what you mean for "get the emulator to not recreate on rotation"(Sorry My English is not good).
[...]

PR #549, opened 2012-07-12
Preserve tab behavior onOrientationChange (issue #365)

yigit:
This commit fixes tab behavior bug onOrientationChange when Activity handles orientation changes (android:configChanges="orientation").

Probably, this is not the proper implementation because it re-creates tab layout but for our case, it solves the problem.
Tested on Android 2.3.4 and Android 4.0.4.

### DIFF
--- a/library/src/com/actionbarsherlock/internal/app/ActionBarImpl.java
+++ b/library/src/com/actionbarsherlock/internal/app/ActionBarImpl.java
@@ -187,6 +187,7 @@ public class ActionBarImpl extends ActionBar {
                 mContextView.onConfigurationChanged(newConfig);
             }
         }
+        recreateTabs();
     }
 
     private void setHasEmbeddedTabs(boolean hasEmbeddedTabs) {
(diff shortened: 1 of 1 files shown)


Unl0gic:
I've tested these changes and they indeed solve the problem with tab rotation when the Activity is configured using android:configChanges="orientation|keyboardHidden" in the manifest.

SimonVT:
Tabs aren't the only thing affected by runtime changes, which is why this won't be merged. That, and handling runtime changes is horrible.

Unl0gic:
Okay, I understand that the fix may have side effects and cannot be merged. The bug however is valid and causes tabs to become unresponsive when orientation changes on Android 2.x (haven't tested 3.x).

The native ActionBar tabs on Android 4.x handles the rotation just fine when android:configChanges="orientation|keyboardHidden" is set on the activity.

Edit: Ignore this comment, I found Jake's reply on the bug here https://github.com/JakeWharton/ActionBarSherlock/issues/279

PR #558, opened 2012-07-20
force close fix

nschwermann:
fixes force close nullpointerexception caused by using a menu button in a fragment to replace the same fragment therefore removing the button.


JakeWharton:
There are way too many changes in the pull request. You are changing this that are totally unrelated to the fix. Please resubmit with only the part of the code that you think has an error.

Issue #564, opened 2012-08-02
ArrayIndexOutOfBoundsException occurs randomly when starting an Activity with ABS


atermenji:
I am writing an application with several SherlockFragmentActivities for open-source [Catroid](https://github.com/Catrobat/Catroid) project 
There is a test project for this application to test UI. It has nearly 150 test cases which are using Robotium to test the behavior. The test project is running daily on Jenkins.
The problem is that sometimes random ArrayIndexOutOfBoundsExceptions occur in different tests when running test project on Jenkins. The stack trace is always the same:

`08-02 01:31:53.480 E/AndroidRuntime( 8486): FATAL EXCEPTION: main
08-02 01:31:53.480 E/AndroidRuntime( 8486): android.view.InflateException: Binary XML file line #17: Error inflating class <unknown>
08-02 01:31:53.480 E/AndroidRuntime( 8486):     at android.view.LayoutInflater.createView(LayoutInflater.java:518)
08-02 01:31:53.480 E/AndroidRuntime( 8486):     at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:570)
08-02 01:31:53.480 E/AndroidRuntime( 8486):     at android.view.LayoutInflater.inflate(LayoutInflater.java:386)
08-02 01:31:53.480 E/AndroidRuntime( 8486):     at android.view.LayoutInflater.inflate(LayoutInflater.java:320)
08-02 01:31:53.480 E/AndroidRuntime( 8486):     at com.actionbarsherlock.internal.view.menu.BaseMenuPresenter.createItemView(BaseMenuPresenter.java:155)
08-02 01:31:53.480 E/AndroidRuntime( 8486):     at com.actionbarsherlock.internal.view.menu.BaseMenuPresenter.getItemView(BaseMenuPresenter.java:174)
08-02 01:31:53.480 E/AndroidRuntime( 8486):     at com.actionbarsherlock.internal.view.menu.ActionMenuPresenter.getItemView(ActionMenuPresenter.java:187)
08-02 01:31:53.480 E/AndroidRuntime( 8486):     at com.actionbarsherlock.internal.view.menu.ActionMenuPresenter.flagActionItems(ActionMenuPresenter.java:460)
08-02 01:31:53.480 E/AndroidRuntime( 8486):     at com.actionbarsherlock.internal.view.menu.MenuBuilder.flagActionItems(MenuBuilder.java:1048)
[...]

PR #569, opened 2012-08-06
Fix for issue #564

atermenji:
Issue appeared when a lot of UI tests which used Robotium were running one after another on Jenkins. When getResources_getBoolean() method from ResourcesCompat.java was surrounded with try catch block, issue disappeared.  
Not sure what was causing that ArrayIndexOfBoundsException. Probably internal Robotium logic of initializing and finishing test activities. Sometimes they are not finished correctly and keep hanging in memory.
Or, probably, ActionBarSherlock's options menu was not initializing correctly in some cases.


aried3r:
I'm also very interested in this ticket. Two projects of ours now fail tests on our CI server because of this. Will this be resolved in 4.2.0?

JakeWharton:
This is not a viable solution to a problem which is not actually present in ActionBarSherlock itself. If this method is failing then there is a problem elsewhere in Android. Falling back to the resources on pre-API 13 devices will almost always return the wrong result for the desired value.

PR #574, opened 2012-08-13
Fix: bad version range for maven-replace-plugin Eclipse lifecycle exclusion.

rgladwell:
Just a small fix: Eclipse lifecycle mapping exclusion does not match version of maven-replace-plugin causing build errors in Eclipse m2e.


JakeWharton:
I merged this to `dev` a few days ago. Will be released with the next version of the library (hopefully in about two weeks)

rgladwell:
Thanks for that @JakeWharton

rgladwell:
@JakeWharton is there a reason you didn't also merge commit 21b8777?

JakeWharton:
Is that a necessary part of the fix? I didn't realize the `groupId` was optional in child modules if it matched the parent. If that is indeed the case I'd like to remove it from all the other children as well.

rgladwell:
@JakeWharton no, not required. i can create a new pull request that removes all the redundant declarations in the child poms?

JakeWharton:
Nope. All over it. Thanks for the tip!

Issue #565, opened 2012-08-03
Sherlock's AutoCompleteTextView style doesn't work


gabrielittner:
I use a theme with Theme.Sherlock.Light as parent. As log as I don't define android:autoCompleteTextViewStyle or use @style/Widget.Sherlock.Light.AutoCompleteTextView the AutoCompleteTextView has no background, paddings and so on. Using @android:style/Widget.AutoCompleteTextView fixes this.
The difference between the Widget.AutocompleteTextView and Sherlock's style is that the stock style has Widget.EditText as parent or defines @android:drawable/edit_text as background (gingebread and earlier).

Gingerbread

```
<style name="Widget.AutoCompleteTextView">
    <item name="android:focusable">true</item>
    <item name="android:focusableInTouchMode">true</item>
    <item name="android:clickable">true</item>
    <item name="android:background">@android:drawable/edit_text</item>
    <item name="android:completionHintView">@android:layout/simple_dropdown_hint</item>
    <item name="android:textAppearance">?android:attr/textAppearanceMediumInverse</item>
    <item name="android:gravity">center_vertical</item>
    <item name="android:completionThreshold">2</item>
    <item name="android:dropDownSelector">@android:drawable/list_selector_background</item>
    <item name="android:popupBackground">@android:drawable/spinner_dropdown_background</item>
    <item name="android:dropDownVerticalOffset">-6dip</item>
    <item name="android:dropDownHorizontalOffset">0dip</item>
    <item name="android:dropDownWidth">wrap_content</item>
</style>
```

ICS

```
<style name="Widget.AutoCompleteTextView" parent="Widget.EditText">
    <item name="android:completionHintView">@android:layout/simple_dropdown_hint</item>
    <item name="android:completionThreshold">2</item>
    <item name="android:dropDownSelector">@android:drawable/list_selector_background</item>
    <item name="android:popupBackground">@android:drawable/spinner_dropdown_background</item>
    <item name="android:dropDownVerticalOffset">-6dip</item>
    <item name="android:dropDownHorizontalOffset">0dip</item>
    <item name="android:dropDownWidth">wrap_content</item>
</style>
```

PR #578, opened 2012-08-14
fix issue #565

gabrielittner:
Use "@android:style/Widget.EditText" instead of Widget as parent for AutoCompleteTextView style, to solve issue #565. As pointed out there in ICS the stock style uses it too as parent. I tested in on Gingerbread, Honeycomb and ICS (level 10,13,15) and it works.

### DIFF
--- a/library/res/values-v14/abs__styles.xml
+++ b/library/res/values-v14/abs__styles.xml
@@ -85,6 +85,11 @@
     <style name="Widget.Sherlock.Light.ProgressBar.Horizontal" parent="android:Widget.Holo.Light.ProgressBar.Horizontal">
     </style>
 
+    <style name="Widget.Sherlock.AutoCompleteTextView" parent="android:Widget.Holo.AutoCompleteTextView">
+    </style>
+    <style name="Widget.Sherlock.Light.AutoCompleteTextView" parent="android:Widget.Holo.Light.AutoCompleteTextView">
+    </style>
+
     <style name="TextAppearance.Sherlock.Widget.ActionBar.Menu" parent="android:TextAppearance.Holo.Widget.ActionBar.Menu">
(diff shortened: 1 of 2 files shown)


JakeWharton:
I would prefer to pull in the explicit styles required and avoid the dependency on a platform style. OEMs love to destroy platform styles on pre-3.0 with their own ugliness.

gabrielittner:
If an OEM changes the EditText design on pre-3.0  imo the changed style should be used. Otherwise an AutoCompleteTextView would look different than all other EditTexts on this device. I think this wouldn't be very good.

For 3.0 and later I will make android:Widget.Holo.EditText parent to avoid OEM changes to the general style.

gabrielittner:
Hey Jake is the current pull request ok for you or should I pull in the EditText style and the EditText background/color for the theme? I would leave it this way, so all EditText look consistent across one device.

JakeWharton:
ActionBarSherlock's goal is to ensure the action bar looks the same across all devices, not to make sure controls look the same as the device default. Most manufacturers destroy the built-in themes which is why I do not want to rely on their styles--especially not for something as essential as the `SearchView` component.

gabrielittner:
I got an alternative idea. How about using a new attribute for the AutoCompleteTextView style, which is applied at the SearchView layout file? This way it wouldn't interfere with non ActionBar stuff. Otherwise we need to pull in at least all the Gingerbread and Holo EditText backgrounds (in theory Froyo also has different ones), so that we don't change the platform look and feel.

gabrielittner:
Sorry pressed the wrong button again.

JakeWharton:
That sounds like a more reasonable approach.

gabrielittner:
New pull request #597

PR #581, opened 2012-08-17
Update android support package to r10

chalup:


willlunniss:
Just a warning that the new NoticationCompat implementation (the only benifit of rev 10), tries to load the ICS implementation for android 3.2, which then causes a FC due to NoSuchMethodError http://code.google.com/p/android/issues/detail?id=36359 so its not really worth updating.

chalup:
The previous version is r7 and the change list of r8 and r9 looks quite useful, for example "Fixed numerous bugs for ViewPager" (full list on http://developer.android.com/tools/extras/support-library.html).

If the r10 is in fact screwed up, let's bump the support package to r9 (I'll update this pull request as soon as I find the jar).

JakeWharton:
We can't upgrade the library until a newer version of the support library makes its way into Maven central. If you need a newer version and you compile with ant (and since Android lacks any semblance of real dependency management), I suggest you maintain your own fork with whatever revision you desire.

PAStheLoD:
Hello, and sorry for digging up these bones, but I'm rather confused about ... why include the support-v4 library at all, if it's referenced in the POM? On the other hand, if you include it, why keep it to the same revision as mvn-central has? Or, maybe for the ant-only users? But they already have to do a bit of android update lib-project.

JakeWharton:
I mean, I guess that there's no real reason to keep them in sync. And yes, it's for the ant/Eclipse users.

PR #583, opened 2012-08-18
updated android support binary to v10

fernandogcat:
just an update for the binary. I think it was at v7 and it has now several updates 

you can see revision history here http://developer.android.com/tools/extras/support-library.html

Thanks for your work!


fernandogcat:
I have missed update demos as well and I'm not sure if it's working properly with the updated revision. I'm gonna close and check better before. Sorry!

SimonVT:
Please see #581 for why a pull request like this will not be accepted.

Issue #540, opened 2012-07-04
Caught ClassNotFoundException caused by android:actionProviderClass?


MalcolmEvershed:
I ran the com.actionbarsherlock.sample.styled app on the Honeycomb MR2 emulator (API level 13) and I saw logcat output like:

<pre>
07-04 02:05:04.361: W/Resources(385): Converting to string: TypedValue{t=0x1/d=0x103011c a=-1 r=0x103011c}
07-04 02:05:04.371: W/MenuInflater(385): Ignoring attribute 'actionProviderClass'. Action view already specified.
07-04 02:05:04.494: W/Resources(385): Converting to string: TypedValue{t=0x1/d=0x103011c a=-1 r=0x103011c}
07-04 02:05:04.501: W/MenuInflater(385): Cannot instantiate class: @16974108
07-04 02:05:04.501: W/MenuInflater(385): java.lang.ClassNotFoundException: @16974108 in loader dalvik.system.PathClassLoader[/data/app/com.actionbarsherlock.sample.styled-2.apk]
07-04 02:05:04.501: W/MenuInflater(385):    at dalvik.system.PathClassLoader.findClass(PathClassLoader.java:251)
07-04 02:05:04.501: W/MenuInflater(385):    at java.lang.ClassLoader.loadClass(ClassLoader.java:540)
07-04 02:05:04.501: W/MenuInflater(385):    at java.lang.ClassLoader.loadClass(ClassLoader.java:500)
07-04 02:05:04.501: W/MenuInflater(385):    at com.actionbarsherlock.view.MenuInflater$MenuState.newInstance(MenuInflater.java:463)
07-04 02:05:04.501: W/MenuInflater(385):    at com.actionbarsherlock.view.MenuInflater$MenuState.readItem(MenuInflater.java:366)
07-04 02:05:04.501: W/MenuInflater(385):    at com.actionbarsherlock.view.MenuInflater.parseMenu(MenuInflater.java:147)
07-04 02:05:04.501: W/MenuInflater(385):    at com.actionbarsherlock.view.MenuInflater.inflate(MenuInflater.java:97)
07-04 02:05:04.501: W/MenuInflater(385):    at com.actionbarsherlock.sample.styled.MainActivity.onCreateOptionsMenu(MainActivity.java:88)
[...]

PR #587, opened 2012-08-24
Fix issue #540

MichaelRocks:
The issue can be reproduced when running `sample-styled` on Android 3.2 emulator. After this fix no warnings appear in logcat while styled attributes still work fine.

I've implemented `Context.obtainStyledAttributes()` method using `Resource.obtainAttributes()` and `Theme.resolveAttribute()` methods. You can find the implementation in `ContextCompat` and `TypedArrayCompat` classes. It also required me to port some part of an internal `XmlUtils` class. Moreover, I've added some styled attributes to `sample-styled`.

Hope this will help you!


JakeWharton:
Fixed by #589. Thanks for the contribution though!

PR #606, opened 2012-09-07
Updated android-support-v4.jar to revision 10.

yanchenko:
For the poor non-Maven users.
http://developer.android.com/tools/extras/support-library.html


JakeWharton:
See #581 for why this can't be merged. Plus you left out all the sample apps.

PR #607, opened 2012-09-07
Update library/res/layout/sherlock_spinner_dropdown_item.xml

notnil:
This change made my ActionBar Drop-Down Navigation look the same as native Drop-Down Navigation.


notnil:
The difference was on Android 2.3.

SimonVT:
This attribute is already defined in the default style. I see no difference between native implementation and when running on 2.3.

notnil:
Before Addition:



After Addition:

SimonVT:
If you're redefining any theme attribute, make sure you extend the default styles. I've loaded up the samples in a 2.3 emulator and I'm not seeing this behavior.

notnil:
K.  I am just using it in mine so I figured I'd see if it would be useful to others.

PR #615, opened 2012-09-17
Fix using android:dropDownHorizontalOffset by IcsSpinner from custom styles

Alexious-sh:
android:dropDownHorizontalOffset  style attribute doesn't work, because it was overwriten by paddings offset in DropdownPopup.show.
- Correct using android:dropDownSelector by IcsSpinner

### DIFF
--- a/library/pom.xml
+++ b/library/pom.xml
@@ -11,7 +11,7 @@
 	<parent>
 		<groupId>com.actionbarsherlock</groupId>
 		<artifactId>parent</artifactId>
-		<version>4.0.3-SNAPSHOT</version>
+		<version>4.1.0</version>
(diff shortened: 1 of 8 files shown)


Alexious-sh:
Sorry, I novice on github, but what's wrong?

JakeWharton:
You are using tabs instead of spaces which is failing the build.

Alexious-sh:
I don't understand what it want from me. I replaced tabs with spaces, but still "build failure". Locally library compiles without any error with my changes.
May be I need to create new pull request with all changes in one commit?

SimonVT:
You fixed some checkstyle errors in the second commit, there are still some in the first. Also, please squash it into a single commit.

PR #622, opened 2012-09-18
Save action view state.

### LINKED ISSUE #566: Collapsible EditText not restoring instance state on rotation
(**NOTE**:  (Mostly) minimal reproduction code here: https://github.com/MichaelSims/CollapsibleActionItemRotationTest/tree/action-bar-sherlock)

Observed on HTC Thunderbolt (2.3.4) using ActionBarSherlock version 4.1.0
##### Steps to reproduce
- Add a collapsible item to the action bar using `android:showAsAction="always|collapseActionView"` and `android:actionLayout` pointing to a layout file containing an EditText
- Expand the action item, enter some text into the EditText field.
- Rotate the device
##### Expected behavior

After rotation, the collapsed/expanded state should be restored, and the EditText action item should remain expanded.  State associated with the EditText (contents, cursor position, etc.) should be maintained.
##### Actual behavior

The collapsed/expanded state is reset, and the collapsible item is collapsed.  Upon expanding the item again, the state of the EditText is gone (empty).

Note that this behavior was also observed in the sample-demos app (Collapsible Action Items sample).  The framework implementation exhibits the expected behavior above on Asus Nexus 7" tablet running Android 4.1.1

Original google groups discussion: https://groups.google.com/forum/#!topic/actionbarsherlock/kZlRLtdZ_4E/discussion

SimonVT:
Closes #566.

### DIFF
--- a/library/src/com/actionbarsherlock/ActionBarSherlock.java
+++ b/library/src/com/actionbarsherlock/ActionBarSherlock.java
@@ -537,6 +537,9 @@ public abstract class ActionBarSherlock {
      */
     public void dispatchDestroy() {}
 
+    public void dispatchSaveInstanceState(Bundle outState) {}
+
+    public void dispatchRestoreInstanceState(Bundle savedInstanceState) {}
 
     ///////////////////////////////////////////////////////////////////////////
     ///////////////////////////////////////////////////////////////////////////
--- a/library/src/com/actionbarsherlock/app/SherlockActivity.java
+++ b/library/src/com/actionbarsherlock/app/SherlockActivity.java
@@ -116,6 +116,17 @@ public abstract class SherlockActivity extends Activity implements OnCreatePanel
         return super.dispatchKeyEvent(event);
     }
 
+    @Override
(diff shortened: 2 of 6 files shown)


SimonVT:
Not testable on the 2.1 emulator: http://code.google.com/p/android/issues/detail?id=2423
Edit: Ignore this message, it works now.

library/src/com/actionbarsherlock/ActionBarSherlock.java
```diff
@@ -537,6 +537,9 @@ public void dispatchPanelClosed(int featureId, android.view.Menu menu) {}
      */
     public void dispatchDestroy() {}
 
+    public abstract Bundle saveHierarchyState();
+
+    public abstract void restoreHierarchyState(Bundle savedInstanceState);
```
JakeWharton: These both should be `void` and prefixed with 'dispatch'. This way all of the logic lives inside of the `ActionBarSherlockCompat` class and reduces duplication in activities.

JakeWharton:
Yeah, can you rename it to match?

PR #625, opened 2012-09-19
Fix setting dropDownSelector and dropDownHorizontalOffset to IcsSpinner from custom style

Alexious-sh:


JakeWharton:
Can you detail more about why this is needed? I don't remember having ever seen an issue around this.

Alexious-sh:
If I remember correct, dropDownSelector style attribute can be used to define drawable, displayed as item background, when it selected by arrow keys. It value ignored by IcsSpinner, but works in real Holo Spinner.

dropDownHorizontalOffset ignored for now too.
When I push this pool request, I develop custom style for spinner, using IcsSpinner from ABS. And it looked better with little bigger margin from left of main element.

PR #631, opened 2012-09-22
Library relocation

rgladwell:
The Maven relocation feature allows us to move the ActionBarSherlock artifactId to a more meaningful string without breaking people's builds:

http://maven.apache.org/guides/mini/guide-relocation.html


rgladwell:
Sorry, wrong branch.

PR #632, opened 2012-09-22
Relocate ActionBarSherlock library -> actionbarsherlock-library

rgladwell:
The Maven relocation feature allows us to move the ActionBarSherlock artifactId to a more meaningful string without breaking people's builds:

http://maven.apache.org/guides/mini/guide-relocation.html


JakeWharton:
Can't be auto-merged anymore. I'd prefer to do this without renaming the folder itself. I will implement for 4.2.

rgladwell:
Fair enough, when do you think you'll be releasing 4.2?

JakeWharton:
It's about a week away. Hopefully sooner but it all depends on my workload
this coming week.

rgladwell:
Thanks for the feedback.

PR #645, opened 2012-10-02
Don't call onOptionsItemSelected for fragments in backstack.

Vandalko:
In some rare cases(with FragmentTransaction.replace() method) fragment could be in backstack but mHidden flag is still false. I think that this is support library bug but anyway there is easy way to fix that.

P.S. ActionBar changes are optional and used only to extract duplicated code.


library/src/android/support/v4/app/Watson.java
```diff
@@ -123,7 +123,7 @@ public boolean onMenuItemSelected(int featureId, MenuItem item) {
             if (mFragments.mAdded != null) {
                 for (int i = 0; i < mFragments.mAdded.size(); i++) {
                     Fragment f = mFragments.mAdded.get(i);
-                    if (f != null && !f.mHidden && f.mHasMenu && f.mMenuVisible && f instanceof OnOptionsItemSelectedListener) {
+                    if (f != null && !f.mHidden && !f.isInBackStack() && f.mHasMenu && f.mMenuVisible && f instanceof OnOptionsItemSelectedListener) {
```
Vandalko, line 126: It's very rare case but i'm working hard with FragmentTransaction.replace() method which is a little buggy.

library/src/com/actionbarsherlock/app/ActionBar.java
```diff
@@ -890,6 +892,14 @@ public void setHomeButtonEnabled(boolean enabled) { }
         public void onTabReselected(Tab tab, FragmentTransaction ft);
     }
 
+    protected FragmentTransaction beginTabChangeTransaction(final Activity activity) {
```
Vandalko, line 895: just extracted duplicated code

PR #653, opened 2012-10-11
Fixed the suggestions in the compatibility SearchView.

mattkranzler5:
...ionsAdapter to use the SearchableInfo object to build the uri to query the ContentResolver with. Added missing drawables and defined the necessary attributes for the suggestions in the dropdown to work properly.


DSteve595:
Yes!

ngdinhtoan:
Could you also fix suggestion behavior when user click on search view?
Currently, when we click on search view, it will hide suggestion list view. When we click again, it don't show suggestion again.
It should always show suggestion list when user click on search view (even it's focused) like native search view.

Code:
Call onQueryRefine(mQueryTextView.getText()); on click on mQueryTextView event.

(I'm not sure it is the best way, but atleast it works for me)

JakeWharton:
Merged to `dev`. Thanks!

tprochazka:
Is this in 4.3.0 already?

tprochazka:
I will reply to myself. Yer it is supported from 4.3.0 ;-)

PR #672, opened 2012-10-22
fix will call search twice when use searchview

googolmo:
fix will call search twice when use searchview


JakeWharton:
The native `SearchView` does not have this check.

JakeWharton:
I cannot change `SearchView` behavior unless an associated change is accepted into AOSP for `android.view.SearchView`. Once that is done I will modify its behavior (and we can re-open this issue).

PR #675, opened 2012-10-22
If you disable the action, this icon will become transparent. 

yuki2006:


JakeWharton:
You should use a state-list drawable for this functionality. Provide a resource for `state_enable="false"` that is different from the normal resource.

PR #684, opened 2012-11-01
Added support for transparent holo themes which were missing

savvasdalkitsis:
The only available translucent theme on Android (and therefore in ActionbarSherlock) is :

```
@android:style/Theme.Translucent.NoTitleBar
```

The problem with this theme is that any component added to an Activity which has this theme, will be styled with a Gingerbread UI.

This commit adds 3 themes to ActionbarSherlock :

```
Theme.Sherlock.Translucent
Theme.Sherlock.Translucent.NoTitleBar
Theme.Sherlock.Translucent.NoTitleBar.Fullscreen
```

### DIFF
--- a/library/res/values/abs__themes.xml
+++ b/library/res/values/abs__themes.xml
@@ -182,6 +182,23 @@
         <item name="windowActionBar">false</item>
         <item name="windowNoTitle">true</item>
     </style>
+    <style name="Theme.Sherlock.Translucent">
+        <item name="android:windowBackground">@android:color/transparent</item>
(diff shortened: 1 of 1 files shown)


madlymad:
+1

JakeWharton:
I don't think these make sense to add. The `Theme.Sherlock.*` themes are analogous to `Theme.Holo.*` for pre-HC APIs. We cannot provide every variant of themes since it simply would be too many to maintain. If these had been added in API 11 or 14 along with the action bar that would be a different story.

Issue #557, opened 2012-07-19
IllegalStateException at com.actionbarsherlock.widget.ActivityChooserView.onDetachedFromWindow


Digipom:
I've received a few reports of the following exception on the market, using Action Bar Sherlock 88fc341 and the Android Support Library v4, revision 9:

java.lang.IllegalStateException: Observer com.actionbarsherlock.widget.ActivityChooserView$1@405f4658 was not registered.
at android.database.Observable.unregisterObserver(Observable.java:64)
at com.actionbarsherlock.widget.ActivityChooserView.onDetachedFromWindow(Unknown Source)
at android.view.View.dispatchDetachedFromWindow(View.java:6173)
at android.view.ViewGroup.dispatchDetachedFromWindow(ViewGroup.java:1164)
at android.view.ViewGroup.dispatchDetachedFromWindow(ViewGroup.java:1162)
at android.view.ViewGroup.dispatchDetachedFromWindow(ViewGroup.java:1162)
at android.view.ViewGroup.dispatchDetachedFromWindow(ViewGroup.java:1162)
at android.view.ViewGroup.dispatchDetachedFromWindow(ViewGroup.java:1162)
at android.view.ViewGroup.dispatchDetachedFromWindow(ViewGroup.java:1162)
at android.view.ViewGroup.dispatchDetachedFromWindow(ViewGroup.java:1162)
at android.view.ViewRoot.dispatchDetachedFromWindow(ViewRoot.java:1746)
at android.view.ViewRoot.doDie(ViewRoot.java:2757)
at android.view.ViewRoot.die(ViewRoot.java:2727)
at android.view.WindowManagerImpl.removeViewImmediate(WindowManagerImpl.java:218)
at android.view.Window$LocalWindowManager.removeViewImmediate(Window.java:436)
at android.app.ActivityThread.handleDestroyActivity(ActivityThread.java:2824)
at android.app.ActivityThread.handleRelaunchActivity(ActivityThread.java:2929)
at android.app.ActivityThread.access$1600(ActivityThread.java:124)
at android.app.ActivityThread$H.handleMessage(ActivityThread.java:972)
at android.os.Handler.dispatchMessage(Handler.java:99)
at android.os.Looper.loop(Looper.java:130)
at android.app.ActivityThread.main(ActivityThread.java:3806)
at java.lang.reflect.Method.invokeNative(Native Method)
at java.lang.reflect.Method.invoke(Method.java:507)
[...]

Issue #415, opened 2012-03-29
ShareActionProvider does not work if only one intent


theomega:
The `ShareActionProvider` does not work if only a single IntentHandler can handle the intent. In these cases, the button simply does nothing. I tested it on Android 2.1, 2.2. and 4.0, its the same on all versions. You can test this easily by using the emulator. Only the Messaging App in the emulator can handle these events and therefore it does not work. It works fine in the 2.3 emulator for one reason: There is a bluetooth-app which can handle all intents and therefore the bug does not occur.

JakeWharton:
This may be a platform bug. Can you try testing on ICS with the built-in version? (`android.widget.ShareActionProvider`)

JakeWharton:
Confirmed that it's a native bug. Seems to already have been filed as http://b.android.com/25467

Digipom:
I can reproduce this bug too, and the weird thing is it works if you've opened the list once, with at least two intents. 

For example:

Two activities: Open list.
Change intent to one that resolves to one activity.
Open list: Still works.

It doesn't work if you haven't yet opened it.

Seems related to these lines in ActivityChooserView$ActivityChooserViewAdapter:

if (!mShowDefaultActivity && mDataModel.getDefaultActivity() != null) {
                activityCount--;
            }

IF you haven't opened the list yet, activityCount gets subtracted by 1, getCount() returns 0 and the button gets disabled. The list still contains one activity.

Digipom:
So I tried changing this line:

private boolean mShowDefaultActivity;

to:

private boolean mShowDefaultActivity = true;

I was also able to fix the "clickable first time" by adding this to updateAppearance(), inside the following block:

if (activityCount > 0 && historySize > 0) {
...
mAdapter.setShowDefaultActivity(false, false);
}

That fixes the issue for me and I haven't had complaints about this issue on the market since then.

zh163:
Dear, I faced the same problem, but i think it should fixed by the following:
There' two problems in ActivityChooserView.java:
I.

```
public void setActivityChooserModel(ActivityChooserModel dataModel) {
    mAdapter.setDataModel(dataModel);

    int historysize = mAdapter.getHistorySize();
    if (historysize>0) {
        mAdapter.setShowDefaultActivity(false, false);
    } else {
        mAdapter.setShowDefaultActivity(true, true);
    }

    if (isShowingPopup()) {
        dismissPopup();
        showPopup();
    }
}
```

[...]

PR #689, opened 2012-11-04
Work-around for issue #415 (ShareActionProvider does not work if only one intent)

Digipom:
I've used this work-around code for several months in production now, with no issues once the fix for #557 is also applied (I added a couple of extra protections for that issue).


JakeWharton:
You have all kinds of crazyness up in here with subsequent commits...

Make a topic branch and resubmit against `dev` branch please.

Digipom:
I submitted new pull requests https://github.com/JakeWharton/ActionBarSherlock/pull/863 and https://github.com/JakeWharton/ActionBarSherlock/pull/864. Thanks!

PR #694, opened 2012-11-10
changed provider authority so it doesn't conflict with support v4 demos

jgilfelt:
so the fragments sample app and the official support v4 demos app can play nicely together (resolves `INSTALL_FAILED_CONFLICTING_PROVIDER`).


JakeWharton:
Thanks!

PR #714, opened 2012-11-24
Solve tab view measuring issue. Closes #673.

### LINKED ISSUE #673: Stacked tab navigation - view measurement issue
On pre ICS when I have 2 or 4 stacked tabs whose content width should not require scrolling within the tab container, there always seems to be a 1 pixel offset in the internal `IcsLinearLayout` which enables scrolling and the overscroll indicator. There is no problem when there are 3 tabs.

 

This is easily reproducable in the ABS sample app (Feature Toggles or Tab Navigation with 2 tabs).

SimonVT:


library/src/com/actionbarsherlock/internal/widget/TabsLinearLayout.java
```diff
@@ -0,0 +1,113 @@
...
+import android.content.res.TypedArray;
+import android.util.AttributeSet;
+import android.view.View;
+import android.widget.LinearLayout;
+
+public class TabsLinearLayout extends IcsLinearLayout {
+    private static final int[] R_styleable_LinearLayout = new int[] {
+            /* 1 */ android.R.attr.measureWithLargestChild,
+    };
+    private static final int LinearLayout_measureWithLargestChild = 1;
```
JakeWharton: Should be 0. The comment too.

library/src/com/actionbarsherlock/internal/widget/TabsLinearLayout.java
```diff
@@ -0,0 +1,113 @@
...
+        super.onMeasure(widthMeasureSpec, heightMeasureSpec);
+
+        final int childCount = getChildCount();
+        if (childCount <= 2) return;
+
+        final int mode = MeasureSpec.getMode(widthMeasureSpec);
+        if (mUseLargestChild && mode == MeasureSpec.UNSPECIFIED) {
+            final int orientation = getOrientation();
+            if (orientation == HORIZONTAL) {
+                    useLargestChildHorizontal();
```
JakeWharton: Odd indent.

PR #736, opened 2012-12-11
Adding TargetApi annotations to stop lint from complaining

jberkel:


library/pom.xml
```diff
@@ -24,7 +24,11 @@
 			<groupId>com.google.android</groupId>
 			<artifactId>support-v4</artifactId>
 		</dependency>
-
+		<dependency>
+			<groupId>com.google.android</groupId>
+			<artifactId>annotations</artifactId>
+			<version>4.1.1.4</version>
```
JakeWharton, line 30: Version numbers should only be specified in the root `pom.xml`.

library/src/com/actionbarsherlock/app/SherlockActivity.java
```diff
@@ -18,6 +19,7 @@
 import com.actionbarsherlock.view.MenuInflater;
 import com.actionbarsherlock.view.MenuItem;
 
+@TargetApi(11)
```
JakeWharton, line 22: This is wrong since it works on API 7+

JakeWharton:
Except for the native wrappers, all of these annotations are incorrect. They are used and working on API 7+.

library/src/com/actionbarsherlock/app/SherlockActivity.java
jberkel, line 22: it does, but it overrides an API level 11 method:

http://developer.android.com/reference/android/app/Activity.html#invalidateOptionsMenu%28%29

i just followed lint's output. without these annotation any project embedding ABS will fail linting, even though in practice it will work on 7+. that's what the `@TargetApi` annotations are for (telling lint that it's actually ok and temporarily changing the sdk level)

JakeWharton, line 22: Then the declaration should be on the method, not the entire class.

SimonVT, line 22: Blindly adding annotations to pass lint checks is just wrong, and defeats the entire purpose of lint. You might as well just not run it if this is your approach. A lint.xml that simply ignores the api level check would be more correct.

Also, that is not what @TargetApi is for. The proper way of using @TargetApi is to have one method that calls through to newer APIs annotated with an appropriate API level, and a second method exposed that only uses the new method when the API level allows for it. You can split this up further into static classes, if you really want to separate your compatibility code. You do not annotate an entire class just because it overrides/calls a single method that was introduced in a later API level.

In this case, just annotating e.g. invalidateOptionsMenu() would also be wrong. The method works perfectly fine on API level 7+. It's basically a design choice to allow the use of invalidateOptionsMenu() on older platforms, since it requires less changes when migrating to/from ABS (pretty much everything exposed by ABS has this in mind).

tl;dr if you want to go through ABS and annotate it with @TargetApi, do it properly. Don't just blindly annotate classes to pass the tests. If you just want to shut it up, use lint.xml.

jberkel, line 22: guess i was a bit too eager there. i went through it again to just annotate the specific methods (soundcloud/ActionBarSherlock@23d3cea1dbd3a213cbfdb6c5554499dae26c2b97). 

@SimonVT didn't find a way to exclude NewApi errors for certain packages, maybe this would be the preferable approach. On the other hand, having ABS checked by lint itself is not a bad idea either.

I guess the quick solution is just to exclude ABS from lint, since it's known to work on 7+.

JakeWharton, line 22: The problem is that a lot of the times what we do is a bit complex for lint. It may be useful to just suppress a few overall. @SimonVT is right about the `invalidateOptionsMenu` call, we override it so that it will technically work for all API levels. I find lint highly useful for applications but it's hard to shove into a library where we rely on certain behavior.

SimonVT, line 22: You can suppress the NewApi check for the entire library with lint.xml. It would also be more correct than your commit, since most of the methods you annotated works perfectly fine on older platforms. They "just happen to" (read: on purpose) have the same name as new APIs.

jberkel, line 22: ok, that's a much better solution. only (minor) problem is that all classes have to be listed since lint doesn't accept recursive ignores (example here for reference: http://stackoverflow.com/questions/13516327/writing-custom-lint-xml)

SimonVT, line 22:
``` xml
<issue id="NewApi" severity="ignore" />
```

to ignore it in the entire library, not just single classes.

SimonVT, line 22: Okay, some github formatting messed that up it seems.. Can't be bothered to try and fix it from my phone, I'll pastebin something tomorrow if you haven't figured it out yet.

jberkel, line 22: that would be great, thanks, can't find any documentation on this feature.

jberkel, line 22: btw, what you posted just ignores everything. i still want my own code to be flagged as warnings.

SimonVT, line 22: You put the lint.xml that ignores NewApi completely in the ABS library
folder, not your own project.

jberkel, line 22: i build my project using maven so there is no ABS library folder per se (get recreated for each build) so this might not be applicable. thanks for your help anyway!

PR #745, opened 2012-12-18
Fix CompactMenu background

DDRBoxman:


SimonVT:
The framework doesn't do this, why should we?
Also, the first two attributes are already inherited from Theme.Sherlock.Light.

DDRBoxman:
Ah, that makes more sense. I was looking at the menu on a 2.3 device and it had the incorrect background color, not text color.
Should have checked on the Nexus S first.

DDRBoxman:
I noticed that the background value for the CompactMenu themes is set to null. Would it have any adverse effects if those were set to the theme background colors?

SimonVT:
It has to behave and look as its native counterpart (on ICS). If the native themes don't set these attributes, I'm not sure this is the right approach.

DDRBoxman:
I did some research into the panel menu background and this seems to make the most sense.

DDRBoxman:
In that case it seems that the only way to style that is to bypass the native menu completely and use a custom panel in it's place.

Issue #746, opened 2012-12-19
homeAsUp Not Enabling Home Action Item


JakeWharton:
This behavior changed in API 16 but we can fix it for all API levels easily. Read the attribute on 14+ and manually call the API. On pre-14 just "fix the glitch".

SimonVT:
What changed exactly?

JakeWharton:
When setting homeAsUp in the theme's displayOptions the home action item was NOT enabled on pre-16. Starting with API 16 they now correctly enable the item when set in the theme.

We should be able to correct this behavior ourselves.

PR #754, opened 2012-12-21
Fix for #746 (homeAsUp) issue

matthewmichihara:


JakeWharton:
Don't we also need to do this for the compatible action bar?

JakeWharton:
Needs to be against `dev` branch anyways.

PR #755, opened 2012-12-21
Re-add ABS support for GTV menu button

roostr:
This patch restores functionality to show overflow menu when the menu button is pressed, but only on GTV systems.


willlunniss:
Are you considering accepting this, or just leaving it open for people to find easily?

Up until now I've been using a local copy of ABS with this patch applied, but it would be nice to move to the new aar via gradle at some point soon.

Issue #748, opened 2012-12-19
Using "android.R.id.content" causes "sticky" initial fragment


Adlatus:
I noticed a difference in the behaviour of fragment handling (using FragmentManager/-Transactions) between ABS and the regular Android framework plus android-support-v4.jar library:

I have a (Sherlock)FragmentActivity which hosts two fragments, one of which is full-screen visible at a time. Background style is transparent. When tapping on the activity window, the screen switches between the two fragments.

Before I used ABS, this has worked just fine. Now, since I introduced ABS (because I want to use the ActionBar), the first fragment persists on the screen, and on all subsequent clicks, I see the alternating two fragments **plus the unwanted, persisting first fragment** (at least when I use the transparent theme/style). This is when running on an Android 2.2 phone/emulator, it works fine with Android >= 4.0. Here is my sample code (differences between implementation without and with ABS in brackets [ ]):

``` java
public class MainActivity extends [Sherlock]FragmentActivity implements
        OnClickListener {

    private DemoFragment frag1 = new DemoFragment();
    private DemoFragment frag2 = new DemoFragment();
    private boolean frag1active = true;

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        findViewById(android.R.id.content).setOnClickListener(this);

        frag1.text = "Frag1";
        frag2.text = "Frag2";

        // Activate Fragment1
        FragmentManager fragMgr = this.getSupportFragmentManager();
        FragmentTransaction fragTrans = fragMgr.beginTransaction();
        fragTrans.add(android.R.id.content, frag1);
        fragTrans.commitAllowingStateLoss();
    }

    @Override
    public void onClick(View v) {
        // Switch fragments
        FragmentManager fragMgr = this.getSupportFragmentManager();
        FragmentTransaction fragTrans = fragMgr.beginTransaction();
        fragTrans.replace(android.R.id.content, frag1active ? frag2 : frag1);
        frag1active = !frag1active;
        fragTrans.commitAllowingStateLoss();
    }
}
```

``` java
public class DemoFragment extends [Sherlock]Fragment {

    public String text = "not yet set";

    @Override
    public View onCreateView(LayoutInflater inflater, ViewGroup container,
            Bundle savedInstanceState) {

        Random random = new Random();
        TextView textView = new TextView(getActivity());
        textView.setText(text);
        textView.setTextSize(30);
[...]

PR #760, opened 2012-12-25
Init the action bar during findViewById

### LINKED ISSUE #623: CheckBox/ToggleButton issue when using "android.R.id.content"
This issue is a duplicate of issue 613 (since it was closed I could not reopen it)
I created a view inside a fragment (extending from SherlockFragment)
Fragment calls setHasOptionsMenu(true), sets titlle and setDisplayHomeAsUpEnabled

CheckBox/ToggleButton does not check/uncheck 

Android 2.3.5 and 2.3.6, it works on ics

---

I was able to reproduce this issue:

Layout-----------------------------------------

<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
    android:layout_width="wrap_content"
    android:layout_height="wrap_content"
    android:gravity="top" >

```
<ToggleButton
    android:id="@+id/toggle"
    android:layout_width="wrap_content"
    android:layout_height="wrap_content"
    android:checked="true"
    android:text="test checkbox" />

<CheckBox
    android:id="@+id/chkbox"
    android:layout_width="wrap_content"
    android:layout_height="wrap_content"
    android:checked="true"
    android:text="test checkbox" />
```

</LinearLayout>
# 

Fragment--------------------------------------------------------------------

public class RoundedColourFragment extends SherlockFragment {
    public RoundedColourFragment() {

```
}

@Override
public void onCreate(Bundle savedInstanceState) {
    super.onCreate(savedInstanceState);
}

@Override
public View onCreateView(LayoutInflater inflater, ViewGroup container,
        Bundle savedInstanceState) {
    return inflater.inflate(R.layout.issue613_tb_checkbox, container, false);
}
```

}
# 

Activity---------------------------------------------------------------------------------------
public class Issue613 extends SherlockFragmentActivity {

```
@Override
public void onCreate(Bundle savedInstanceState) {
    super.onCreate(savedInstanceState);
    FragmentManager fm = getSupportFragmentManager();
    FragmentTransaction ft = fm.beginTransaction();
    Fragment newFragment = Fragment.instantiate(this, RoundedColourFragment.class.getName(), null);
    ft.replace(android.R.id.content, newFragment);
    ft.setTransition(FragmentTransaction.TRANSIT_FRAGMENT_OPEN);
    ft.addToBackStack(null);
    ft.commit();
}
```

}
# 
# 

I was able to find a workaround this issue by removing  "android.R.id.content"  and replacing it with an id in layout.

SimonVT:
Ensure that the ActionBar is initialized when findViewById is called.

Closes #623 and #748.

### DIFF
--- a/library/src/com/actionbarsherlock/ActionBarSherlock.java
+++ b/library/src/com/actionbarsherlock/ActionBarSherlock.java
@@ -791,4 +791,9 @@ public abstract class ActionBarSherlock {
      * @see ActionMode
      */
     public abstract ActionMode startActionMode(ActionMode.Callback callback);
+
+    /**
+     * Ensure that the action bar is attached.
+     */
+    public void ensureActionBar() {}
 }
--- a/library/src/com/actionbarsherlock/app/SherlockActivity.java
+++ b/library/src/com/actionbarsherlock/app/SherlockActivity.java
@@ -243,6 +243,12 @@ public abstract class SherlockActivity extends Activity implements OnCreatePanel
         getSherlock().requestFeature((int)featureId);
     }
 
+    @Override
+    public View findViewById(int id) {
+        getSherlock().ensureActionBar();
+        return super.findViewById(id);
+    }
+
 
     ///////////////////////////////////////////////////////////////////////////
     // Progress Indication
--- a/library/src/com/actionbarsherlock/app/SherlockExpandableListActivity.java
+++ b/library/src/com/actionbarsherlock/app/SherlockExpandableListActivity.java
@@ -232,6 +232,12 @@ public abstract class SherlockExpandableListActivity extends ExpandableListActiv
         getSherlock().requestFeature((int)featureId);
     }
 
+    @Override
+    public View findViewById(int id) {
+        getSherlock().ensureActionBar();
+        return super.findViewById(id);
+    }
+
 
     ///////////////////////////////////////////////////////////////////////////
     // Progress Indication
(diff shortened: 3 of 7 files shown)


JakeWharton:
I feel like there was an explicit reason why I chose not to use `onCreate`. Will have to think on it a bit and see if I can remember.

SimonVT:
Maybe it's enough to call installDecor(). But really, only two things can happen. Either setContentView(...) is called, in which case installDecor()/initActionBar() is called during onCreate anyway, or a fragment is added and the view hierarchy is re-ordered, messing it up.

JakeWharton:
I think part of the problem is that you can have code like this:

``` java
super.onCreate(savedInstanceState);
requestWindowFeature(Window.FEATURE_ACTION_BAR_OVERLAY);
requestWindowFeature(Window.FEATURE_INDETERMINATE_PROGRESS);
setContentView(R.layout.whatever);
```

Which would cause the action bar to be initialized too soon and lack a progress bar. You also cannot initialize the just the decor view since then you won't get the overlay either.

SimonVT:
Ah, good point. Hm, well we need some other place to hook in then. findViewById?

JakeWharton:
Yeah that should work. Shame we have to resort to such hacks. When ABS was
integrated with the support lib it was so much easier.

Maybe we could add a hook into the FragmentManager. I'm sure the folk at
Google would allow something simple that was package scoped.

SimonVT:
Alright, I'll get this done tomorrow.

Maybe, it's worth a try at least. Maybe they'll just tell us to override getSupportFragmentManager() (which could also be a possibility).

iNoles:
I am sure of Google would tell you to override package-scoped FragmentManagerImpl for moveToState methods or Fragment class for package-scoped performCreate methods.

JakeWharton:
@iNoles We can't instantiate the instance that's used so that won't work.

PR #772, opened 2013-01-10
 Add a generic logger interface that mimics android.util.Log so another ...

robUx4:
I am working on a generic API for android libraries so it's easier for the lib user to redirect the logs to something else. It's transparent if you don't call LogManager.setLogger()

### DIFF
--- a/actionbarsherlock/src/android/support/v4/app/Watson.java
+++ b/actionbarsherlock/src/android/support/v4/app/Watson.java
@@ -1,11 +1,11 @@
 package android.support.v4.app;
 
-import android.util.Log;
 import android.view.View;
 import android.view.Window;
 import com.actionbarsherlock.ActionBarSherlock.OnCreatePanelMenuListener;
 import com.actionbarsherlock.ActionBarSherlock.OnMenuItemSelectedListener;
 import com.actionbarsherlock.ActionBarSherlock.OnPreparePanelListener;
+import com.actionbarsherlock.log.LogManager;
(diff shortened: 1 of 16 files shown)


JakeWharton:
Why not just use a normal [`Logger`](https://developer.android.com/reference/java/util/logging/Logger.html)?

robUx4:
This is a heavy class to use. An interface that mimics android.util.Log is a lot easier to implement and calls are easy to replace without making any mistakes.

JakeWharton:
I don't see the value-add here. ABS's logging is compiled out due to the `static final` constant that prefixes each call to `Log`. If anything we should just switch it to `BuildConfig.DEBUG` now that it's provided by the tooling to allow enabling and disabling based on your compilation mode rather than flipping a `boolean` in each file.

robUx4:
Not sure having the static boolean centralized will remove the code like it does now (pretty sure it doesn't). 
If you haven't seen my G+ post, the idea is to be able to redirect the logs of any libs to your own logger. The java Logger could fit but is too heavy. A simple interface that just replaces the Android Log class is easy for everyone. 
The question of whether a lib needs some logs known to the upper level is another question that varies depending on each lib.

JakeWharton:
The `BuildConfig` boolean is still a `static final` constant and thus causes `javac` to just throw away code (you can verify with `javap`).

iNoles:
I uses this for my applications, http://code.google.com/p/iosched/source/browse/android/src/com/google/android/apps/iosched/util/LogUtils.java

Issue #791, opened 2013-01-20
SearchView does not fire suggestion.


lity:
When I use "android.widget.SearchView", the search suggestion is fired, "query" method of my  ContentProvider is invoked.

When I change to "com.actionbarsherlock.widget.SearchView", the search suggestion isn't fired.

Other code except "import ...SearchView" are completely the same.

Do you forget this feature?

yaming116:
I also encountered the same problem, even I click SearchButton, onOptionsItemSelected method will not be executed.

lity:
I created a project to show this issue. Please see https://github.com/lity/BugReproduce

SimonVT:
This should be fixed by #653. For future reference, please add test cases to the known bugs project in the dev branch and submit a pull request.

PR #792, opened 2013-01-21
Add test case to reproduce issue #653(#791)

lity:
Search suggestion can be fired if "import android.widget.SearchView" but can't be fired if "import com.actionbarsherlock.widget.SearchView".


lity:
How to make the Travis to retry? I modified an annotation.

SimonVT:
Please add your test case to the dev branch. Travis is broken, dont worry about that.

PR #793, opened 2013-01-21
Add test case for issue #653(#791)

lity:
Search suggestion can be fired if import android.widget.SearchView but can not be fired if com.actionbarsherlock.widget.SearchView


SimonVT:
Search suggestions shows up just fine here.

lity:
I think what shows up fine is the SearchView  of  android official  version. I committed  the  test case with  android official SearchView to tell you the correct behavior of search suggestions. Notice the  import part and manually  switch to ActionBarSherlock version, no  suggestions pop up.

SimonVT:
No, suggestions show up fine with the ABS SearchView.

PR #795, opened 2013-01-21
Add a suggestions adapter to the SearchView demo.

SimonVT:


yaming116:
Why I enter a letter,SearchAutoComplete did not appear android no  filter. How to support voice

PR #800, opened 2013-01-22
Fixes homeAsUp not enabling the home action item. Closes #746.

### LINKED ISSUE #746: homeAsUp Not Enabling Home Action Item

matthewmichihara:
When setting homeAsUp in the theme's display options on pre-16 devices,
the home action item was not being enabled. Fix this by reading the
attribute and manually calling the setHomeButtonEnabled API.


JakeWharton:
Needs rebased off of the latest `dev`. Otherwise looks great!

matthewmichihara:
Thanks Jake!

PR #804, opened 2013-01-23
Fix crash when use IcsSpinner as actionView in ActionBar

TangKe:
MenuInflater throw constructor can not be found exception when IcsSpinner use as android:actionView of a menu item


JakeWharton:
Hmm... `IcsSpinner` is part of the internal API. You _really_ shouldn't be using it yourself. What's your use case for needing it?

TangKe:
OK, got it, Is there any way to provide compatible dropdown feature for lower API level?

JakeWharton:
You can use sub-menus with action items for a normal drop-down. What are you trying to do with the Spinner?

TangKe:
thx, It seems that sub-menus can not been style in android 4.2.1 by override the popMenuStyle, android:popMenuStyle,  dropDownListViewStyle, android:dropDownListViewStyle

PR #808, opened 2013-01-24
SearchView keeps a query on expanded action

ikravchenko:
While working on a project I faced the problem that search term was not saved for SearchView. THen I found in the code, that the text view is automatically cleaned up in onActionExpanded callback. PLease review the change


SimonVT:
SearchView has to behave as its native equivelant. How does the native SearchView behave?

JakeWharton:
Make pull requests against the `dev` branch. And I'm 99% sure that the native `SearchView` has the same behavior.

ikravchenko:
I have checked the native behaviour. You were right. I should have checked it first, sorry. I'll make pull requests to dev thereafter.

SimonVT:
We're not going to change the behavior if it currently matches the native searchview.

PR #821, opened 2013-02-06
Add custom font support to actionbar title

jbarr21:
Add custom font support to actionbar title


SimonVT:
ABS matches the official API. Until this is natively supported, it isn't going to be added to ABS.

Issue #824, opened 2013-02-08
Reattaching same Fragment instance fails


btalb:
I have been struggling with code in my application that creates any given screen based on what screen is requested. While not the most efficient coding solution, it simply un-attaches every fragment from the screen, and re-attaches the correct ones (creating or reusing depending on their existence). The stackoverflow question can be seen here:
http://stackoverflow.com/questions/14764043/fragments-remove-all-fragments-in-a-view

I believe I have narrowed it down to when using ActionBarSherlock (specifically `SherlockFragmentActivity`), reattaching (and specifically `FragmentTransaction.replace()`), fails if the given `Fragment` is the same instance as that of the already attached Fragment.

While it seems like an issue that should just be avoided with a more clever coding structure, I have verified it does not happen in API 11+ with the standard fragment functions, doesn't happen using the Android Compatibility Packages, and is only introduced when moving to `SherlockFragmentActivity` from `Activity` or `FragmentActivity`.

The code I have used that confirmed this is below:

``` java
package com.test.test;

import com.actionbarsherlock.app.SherlockFragmentActivity;

import android.os.Bundle;
import android.content.Context;
import android.graphics.Color;
import android.support.v4.app.Fragment;
import android.support.v4.app.FragmentActivity;
import android.view.Gravity;
import android.view.LayoutInflater;
import android.view.Menu;
import android.view.View;
import android.view.View.OnClickListener;
import android.view.ViewGroup;
import android.widget.Button;
import android.widget.FrameLayout;
import android.widget.LinearLayout;
import android.widget.LinearLayout.LayoutParams;
import android.widget.TextView;

public class MainActivity extends SherlockFragmentActivity {
    // Context
    public static Context context = null;
    myFragment myFrag = null;
    // Views
    FrameLayout fl = null;


    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        context = this;

        LinearLayout ll = new LinearLayout(context);
        ll.setOrientation(LinearLayout.VERTICAL);

        Button b = new Button(context);
        b.setText("Repeat");
        b.setOnClickListener(new OnClickListener() {
            @Override
            public void onClick(View v) {

                // Reattach the same fragment
                getSupportFragmentManager().beginTransaction().replace(fl.getId(), myFrag).commit();

            }
        });

        fl = new FrameLayout(context);
        fl.setId(200);
[...]

PR #825, opened 2013-02-08
Issue 824

btalb:
Known case added for Issue824 as suggested:
https://github.com/JakeWharton/ActionBarSherlock/issues/824


SimonVT:
You forgot to create it against the dev branch.

PR #826, opened 2013-02-08
Create Issue824.java

btalb:
Known case added for Issue824 as suggested:
https://github.com/JakeWharton/ActionBarSherlock/issues/824

Fixed the code comments on the code, aware it was dumb code, but was just a working example thrown together to show the problem.


JakeWharton:
This is using tabs instead of spaces. The manifest also needs updated.

SimonVT:
I see no difference in behavior when extending FragmentActivity instead of SherlockFragmentActivity. Updating the support library solves this however.

PR #838, opened 2013-02-23
YouTube v3 API support

apkelly:
This is my first Pull Request on any project, so forgive me if I've made a mistake. 

This new class extends the YouTubeBaseActivity for the YouTube API v3, so that you can use a YouTubePlayerView in your activity, and still have an ActionBar that works on pre-4.0. Hopefully you'll find this a useful addition to the library.

Regards,

Andy.


actionbarsherlock/pom.xml
```diff
@@ -24,6 +24,14 @@
       <groupId>com.google.android</groupId>
       <artifactId>support-v4</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.android</groupId>
+      <artifactId>youtube</artifactId>
+      <version>v3</version>
+      <scope>system</scope>
+      <systemPath>${project.basedir}/libs/YouTubeAndroidPlayerApi.jar</systemPath>
```
JakeWharton, line 32: Only dependencies in Maven central are allowed, sorry.

JakeWharton:
The best place for something like this is a standalone project.

PR #839, opened 2013-02-25
Fixed MenuWrapper memory leak

Vandalko:
Issue is the same as was with old View.setTag(id, tag) implementation(was discussed [here](http://stackoverflow.com/a/7145210/87383) and [here](https://code.google.com/p/android/issues/detail?id=18273)).

So you have to clean mNativeMap to fit real menu.

P.S. I think it's corner case because it's not a very common case when you have pointers to(or from) menu items from activity, fragments, etc...


Vandalko:
> /home/travis/build.sh: line 107: android: command not found

looks like travis feeling bad :(

SimonVT:
Yes, travis is broken.

SimonVT:
I don't think that second commit (the one named "merged") should be part of the pull request.

SimonVT:
Make sure the your commits pass all checkstyle checks.

Vandalko:
> I don't think that second commit (the one named "merged") should be part of the pull request.

It's because my commit originally applied to master - on "Files changed" tab you can see that only 2 files changed. maybe i should consider reordering this commits.

> Make sure the your commits pass all checkstyle checks.

I still can't get used to Warton's code style... :( i'll fix it in a moment

Vandalko:
ok. cleaned code as much as i can...

actionbarsherlock/src/com/actionbarsherlock/internal/view/menu/MenuWrapper.java
```diff
@@ -108,6 +113,20 @@ public void clear() {
...
     }
 
+    public void invalidate() {
+        if (mNativeMap.isEmpty()) { return; }
+
+        final WeakHashMap<android.view.MenuItem, MenuItem> menuMapCopy = new WeakHashMap<android.view.MenuItem, MenuItem>(mNativeMap.size());
+
+        for (int i = 0; i < mNativeMenu.size(); i++) {
+          final android.view.MenuItem item = mNativeMenu.getItem(i);
+          menuMapCopy.put(item, mNativeMap.get(item));
```
JakeWharton: Needs four-space indent.

actionbarsherlock/src/com/actionbarsherlock/internal/view/menu/MenuWrapper.java
```diff
@@ -108,6 +113,20 @@ public void clear() {
         mNativeMenu.clear();
     }
 
+    public void invalidate() {
+        if (mNativeMap.isEmpty()) { return; }
```
JakeWharton: Either no brackets or put `return` on its own line.

actionbarsherlock/src/com/actionbarsherlock/internal/view/menu/MenuWrapper.java
```diff
@@ -94,11 +94,16 @@ public int addIntentOptions(int groupId, int itemId, int order, ComponentName ca
...
     public void removeItem(int id) {
+        mNativeMap.remove(mNativeMenu.findItem(id));
         mNativeMenu.removeItem(id);
     }
 
     @Override
     public void removeGroup(int groupId) {
+        for (int i = 0; i < mNativeMenu.size(); i++) {
+          final android.view.MenuItem item = mNativeMenu.getItem(i);
+          if (item.getGroupId() == groupId) { mNativeMap.remove(item); }
```
JakeWharton: Needs four-space indent.

actionbarsherlock/src/com/actionbarsherlock/internal/ActionBarSherlockNative.java
```diff
@@ -289,6 +291,7 @@ public void setCustomView(View view) {
         @Override
         public void invalidate() {
             mActionMode.invalidate();
+            if (mMenu != null) { mMenu.invalidate(); }
```
JakeWharton: Either no brackets or put invalidate on its own line.

actionbarsherlock/src/com/actionbarsherlock/internal/ActionBarSherlockNative.java
```diff
@@ -48,6 +48,8 @@ public void dispatchInvalidateOptionsMenu() {
         if (BuildConfig.DEBUG) Log.d(TAG, "[dispatchInvalidateOptionsMenu]");
 
         mActivity.getWindow().invalidatePanelMenu(Window.FEATURE_OPTIONS_PANEL);
+
+        if (mMenu != null) { mMenu.invalidate(); }
```
JakeWharton: Either no brackets or put invalidate on its own line.

Vandalko:
changed my eclipse prefs to fit code style

JakeWharton:
> P.S. I think it's corner case because it's not a very common case when you have pointers to(or from) menu items from activity, fragments, etc...

We do this everywhere. But the assumption is that if `onCreateOptionsMenu` is called again (i.e., you invalidate the options menu) then those references are no longer valid.

Vandalko:
Yeah, but it's very sad when the only reason you have memory leak is because of MenuWrapper's internal logic(like it was with tags in old Android versions). So instead of writing tons of boilerplate code again and again(to clear pointers) in every project, we could just fix MenuWrapper ;) (like they did in Android :) )

PR #843, opened 2013-02-26
Close IcsSpinner DropDown programmatically

hkokocin:
Hi,

 I am using a IcsSpinner within a SlidingMenu in my app. The problem is if I open the IcsSpinner's popup and than close the SlidingMenu programmatically the popup will not be dismissed. I am not sure if this is the best way to solve it but i wrote a little quickfix by providing a method in the IcsSpinner to close the popup programmatically.


SimonVT:
Internal classes are not meant for use outside the library. If you wish to add additional functionality to them, you'll have to maintain your own fork.

PR #848, opened 2013-03-01
Fix static import which throws off Scala builds

appamatto:
For some reason sbt with android-plugin is not able to deal with these static imports.

The error itself is:

```
... SherlockFragment.java:11: value OnCreateOptionsMenuListener is not a member of object com.actionbarsherlock.app.SherlockFragmentActivity
[error] import static com.actionbarsherlock.app.SherlockFragmentActivity.OnCreateOptionsMenuListener;
```

Importing them from Watson.java fixes the problem.

### DIFF
--- a/library/src/com/actionbarsherlock/app/SherlockFragment.java
+++ b/library/src/com/actionbarsherlock/app/SherlockFragment.java
@@ -8,9 +8,9 @@ import com.actionbarsherlock.view.Menu;
 import com.actionbarsherlock.view.MenuInflater;
 import com.actionbarsherlock.view.MenuItem;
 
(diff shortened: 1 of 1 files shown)


JakeWharton:
I'll merge to `dev` sometime today. Did you file a bug against sbt for this?

appamatto:
I filed this bug: https://github.com/sbt/sbt/issues/696

I tried to create a minimal test case but no dice. Have you ever seen anything like this before?

JakeWharton:
Merged to `dev` and fixed the other fragment base class imports. And no, I haven't seen anything like this. Its technically valid Java.

PR #853, opened 2013-03-07
Updating to latest android-support-v4 and Android API level

paulpv:
I don't know if you get these type of Pull Requests often.
My app was using a newer android-support-v4 than ActionBarSherlock.
My app was failing to compile because NotificationCompat.Builder.build() was not defined.
I don't know if this would break any other uses of ABS, but I imagine the risk is either low or they can always go back to an older version of the android-support-v4 if they need to.


SimonVT:
We get these a lot. Duplicate of https://github.com/JakeWharton/ActionBarSherlock/issues/search?q=update+support

JakeWharton:
We cannot update the support library since r7 is the latest in Maven central. If you need a newer version you should replace it in your local copy (or use Maven dependencies to trump it by deploying the newer jar). The library is also compiled with API 14 so the properties file is accurate.

PR #855, opened 2013-03-09
Add required application element.

iamstuffed:
http://developer.android.com/guide/topics/manifest/manifest-element.html


JakeWharton:
Is this really required for library projects?

iamstuffed:
I don't really know if it's absolutely necessary, but according to the documentation linked, it's required. I had issues with the new Gradle based build system validating the AndroidManifest.xml file and failing because it couldn't find the application element. 

Other libraries have an empty application tag.

SimonVT:
Hardly any libraries have an application tag, not even google's own (e.g. support-v7). Sounds more like an issue that should be reported to the tools team.
Either way, pull requests go on the dev branch.

iNoles:
the application tag is required for < activity >

PR #863, opened 2013-03-15
Work-around for issue #415 (ShareActionProvider does not work if only on...

Digipom:
Hello,

I cherry-picked the specific commit as a topic branch off of dev, here is the resubmit. Thanks!

I've used this work-around code for several months in production now, with no issues once the fix for #557 is also applied (I added a couple of extra protections for that issue).


JakeWharton:
Am I reading this correctly that this uses the expanded view for only one share target?

Digipom:
Do you mean the dropdown? Yes, it allows the dropdown when there's only one
share target and that target's not already on the action bar as a separate
icon.

On Fri, Mar 15, 2013 at 2:15 PM, Jake Wharton notifications@github.comwrote:

> Am I reading this correctly that this uses the expanded view for only one
> share target?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/JakeWharton/ActionBarSherlock/pull/863#issuecomment-14976591
> .

## 

## 

Digipom
http://www.digipom.com

JakeWharton:
That's not what I meant, but you clarified my concern.

PR #864, opened 2013-03-15
Additional protections for #557.

Digipom:
This is in conjunction with the pull request for #415


JakeWharton:
When would this ever happen?

iNoles:
http://developer.android.com/reference/android/database/Observable.html#registerObserver%28T%29 If observed registered more than once, it will throws IllegalStateException. There is unregisterObserver for it.

JakeWharton:
I'm aware of that. But when would that ever happen?

Digipom:
I think I saw market reports for this, though I added this in a long time ago so I have no idea how often it happens. On the previous pull request, another commentator was asking when it would be integrated as they seem to have seen the same exceptions in the market.

JakeWharton:
Works for me!

JakeWharton:
Thanks for making these against `dev`, too. Merging manually gets annoying. I **promise** a release is coming soon.

Digipom:
Thanks for all of the hard work, Action Bar Sherlock has been a life saver!

PR #882, opened 2013-03-29
Navigation spinner doesn't properly scroll to the currently selected item if there are many items in the list

androidmoney:
This demonstrates the ABS bug due to the missing setSelection() method in IcsListPopupWindow

### DIFF
--- a/actionbarsherlock-samples/known-bugs/src/com/actionbarsherlock/sample/knownbugs/Issue882.java
+++ b/actionbarsherlock-samples/known-bugs/src/com/actionbarsherlock/sample/knownbugs/Issue882.java
@@ -1,0 +1,131 @@
+package com.actionbarsherlock.sample.knownbugs;
+
+import static android.view.Gravity.CENTER;
+import static android.view.ViewGroup.LayoutParams.MATCH_PARENT;
+import static com.actionbarsherlock.app.ActionBar.NAVIGATION_MODE_TABS;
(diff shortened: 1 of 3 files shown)


actionbarsherlock-samples/known-bugs/src/com/actionbarsherlock/sample/knownbugs/Issue882.java
```diff
@@ -0,0 +1,131 @@
...
+import com.actionbarsherlock.app.ActionBar;
+import com.actionbarsherlock.app.ActionBar.TabListener;
+import com.actionbarsherlock.app.SherlockFragment;
+import com.actionbarsherlock.app.SherlockFragmentActivity;
+
+public class Issue882 extends SherlockFragmentActivity implements
+		OnPageChangeListener, TabListener {
+	private static final int COUNT = 20;
+
+	ViewPager mPager;
```
JakeWharton, line 27: This file looks like it's formatted with tabs or two-space indents. ABS uses 4 spaces. I'll just fix after merging tomorrow.

androidmoney, line 27: Hmmm, I just copy/pasted one of the other classes and made a few changes.  It looked fine in my editor.  Sorry about that.

PR #883, opened 2013-03-29
Fix travis config

hameno:
This should fix the travis builds...I got a similar config working in my projects (e.g. https://github.com/PSDev/LicensesDialog)


JakeWharton:
_/me sips drink waiting for Travis_

hameno:
:D

hameno:
Ahh I should probably change the install command so that it isn't "quiet"...

hameno:
Well, seems to be working...

hameno:
I think that is all I can do for now - not sure why the tests are failing, seems to be a problem with the dependencies and scope maybe?

JakeWharton:
I'm just going to wait out Travis' incompetence rather than hack up some crazy solution. Thanks for trying, though.

PR #894, opened 2013-04-17
Fix compilation with a recent ADT

robUx4:


SimonVT:
Add a lint.xml to the abs project that disables the api checks if you wish to ignore them.

PR #895, opened 2013-04-17
Add Hebrew and Brazilian translations

robUx4:
Although a lot of the other languages could be updated too, they all have "" around the strings which is probably incorrectly generated from a script


SimonVT:
Please re-submit the pull request against the dev branch.

PR #896, opened 2013-04-17
Allow ActionBarSherlock logs to be redirected

robUx4:
Use the Logger interface to log anything from ActionBarSherlock
Use the LogManager class to set a custom Logger to receive all ActionBarSherlock logs


SimonVT:
Duplicate of #772.

PR #897, opened 2013-04-17
Themed dialogs

robUx4:
Add a theme for Dialogs too so that most apps using ActionBarSherlock don't have to do it themselves

Use the ICS style of dialog on ICS and up


SimonVT:
This is outside the scope of ABS. Dialogs don't have an action bar.

Prototik:
By the way: dialogs /can/ have a action bar, this behavior uses in PreferenceScreen in native droid, for example.

brnhffmnn:
Dialogs indeed (can) have an action bar:

https://github.com/android/platform_frameworks_base/blob/master/core/java/android/app/Dialog.java#L266

PR #898, opened 2013-04-17
Compilation

robUx4:
Fix compilation with a recent ADT


JakeWharton:
Fixed with XML config on `dev`.

PR #899, opened 2013-04-17
Languages

robUx4:
Add Hebrew and Brazilian languages
More languages could be fixed, all the strings contain some "" which is probably from an incorrect script


JakeWharton:
The platform wraps all strings in quotes. You can read more about its validity [here](http://developer.android.com/guide/topics/resources/string-resource.html#FormattingAndStyling).

PR #900, opened 2013-04-17
Add a Logger interface to redirect ABS logs if needed

robUx4:


JakeWharton:
This is a fundamental change to the behavior of logging that I don't want to make. ABS allows javac to strip logs at compile time in release builds.

robUx4:
I understand, but the ones that are behind a BuildConfig.DEBUG check will not be executed (and likely not generate code). The others are supposed to be noticed, but if the library user want to generate 0 logs, they can set their own Logger that does nothing.

PR #905, opened 2013-04-19
Fix for NullPointerException while importing into Eclipse

hahnjo:
While importing the new version (4.3.0), I get a NullPointerException because he cannot guess the version for the lifecycle-plugin.
In the old version, there was also a version given (https://github.com/JakeWharton/ActionBarSherlock/blob/4.2.0/library/pom.xml#L124)

### DIFF
--- a/actionbarsherlock/pom.xml
+++ b/actionbarsherlock/pom.xml
@@ -110,6 +110,7 @@
         <plugin>
           <groupId>org.eclipse.m2e</groupId>
           <artifactId>lifecycle-mapping</artifactId>
+          <version>1.0.0</version>
           <configuration>
             <lifecycleMappingMetadata>
               <pluginExecutions>


JakeWharton:
Hmmm it's defined in the base `pom.xml` of the project: https://github.com/JakeWharton/ActionBarSherlock/blob/master/pom.xml#L175-179

Eclipse isn't smart enough to pick that up? I never understood why Eclipse needed this thing anyway. UGH.

hahnjo:
At least, this has worked in my case. OK, I have to admin that I only imported the library module by maven. Would have to test that again.
But the other dependencies also got parsed correctly from the root pom.xml...

JakeWharton:
I don't mind having it there if it just works. Will merge when I get back to a computer.

PR #906, opened 2013-04-19
Removed final from methods that ought to be overridable.

worr:
You declare certain methods in ListActivity and ListFragment that are originally overridable as final. This breaks some applications that rely on overriding these methods.


iNoles:
Working as Designed. You only uses com.actionbarsherlock.view.\* import

worr:
How am I supposed to use ActionBarSherlock in code that overrides these methods then?

SimonVT:
ABS provides its own menu callbacks, all you have to do is change your imports. Look at http://abs.io/usage or the samples.
The native callbacks should not be overridden when using ABS. For starters, menu items added to the native menu will not show up in the compat action bar. Marking these methods final ensures this is caught at compile time

PR #915, opened 2013-04-25
fix NullPointerException when trying to select invalid tab position

GoMino:
Self explanatory, the library should not make the application crash even if application select an invalid tab position


SimonVT:
I get an IndexOutOfBoundsException when selecting an invalid tab position (>= tab count), with both the compat and native implementation. I'm not sure what this really solves.

GoMino:
The crash happen when trying to reselect a tab that has not been added to the actionbar (compat mode) the crash still happen in the native implementation, but I didn't have the time to report the problem yet.

SimonVT:
We can't really fix this in the compat implementation then. You should find a way to handle it in your app.

JakeWharton:
Agreed. In the eyes of the library, crashing is the correct behavior since it matches the platform behavior.

PR #925, opened 2013-05-04
Update abs__strings.xml

rk4an:


JakeWharton:
These files are all created with a script. Is there something specifically wrong with them?

rk4an:
the last line missing, and why we have quote around text? thanks,

abs__searchview_description_voice

JakeWharton:
The platform wraps all strings in quotes which is where these are pulled from. You can read more about its validity [here](http://developer.android.com/guide/topics/resources/string-resource.html#FormattingAndStyling). I'll have to take a look at the missing string tomorrow. If it's missing from every language then re-running the script should be enough to pull it in for all languages.

PR #944, opened 2013-05-20
Update android:versionName to reflect the current version.

aried3r:


JakeWharton:
This is fixed on `dev` and was added as part of the release process. It'll be updated with the next release.

PR #946, opened 2013-05-21
Adds ActionBarDrawerToggle compatible with ActionBarSherlock

ianhanniballake:
Adds ActionBarDrawerToggle helper clas compatible with ActionBarSherlock for use with DrawerLayout added in Support Library revision 13.

Signed-off-by: ianhanniballake ian.hannibal.lake@gmail.com


JakeWharton:
I can't add things like this until the support library is updated in Maven central.

ianhanniballake:
Thanks, I'll create another pull request if that ever happens.

JakeWharton:
Google keeps saying they will... hopefully sooner rather than later.

ianhanniballake:
@JakeWharton - as Google is now publishing Support v13 to Maven as per [this announcement](https://plus.google.com/+AndroidDevelopers/posts/4Yhpn6p9icf), would you consider reopening this pull request?

JakeWharton:
Soon :)

animaonline:
any updates folks?

ankushg:
An update for this would be greatly appreciated!

kihaki:
This is very important for our project, any updates? :D

animaonline:
Here is an alternative way to do it:

@Override
public boolean onOptionsItemSelected(MenuItem item) {

if (item.getItemId() == android.R.id.home) {

if (mDrawerLayout.isDrawerOpen(mDrawerList)) {
mDrawerLayout.closeDrawer(mDrawerList);
} else {
mDrawerLayout.openDrawer(mDrawerList);
}
}

return super.onOptionsItemSelected(item);
}

On Thu, Jun 13, 2013 at 12:22 PM, kihaki notifications@github.com wrote:

> This is very important for our project, any updates? :D
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/JakeWharton/ActionBarSherlock/pull/946#issuecomment-19383439
> .

## 

Roman Alifanov - MCAD - MCTS - MCP - MCT
http://www.animaonline.com

ronaldoeyeem:
I found an issue on this commit. 'onOptionsItemSelected' must return true whenever it consumes the event.

yincrash:
@animaonline that works fine for API 11+ devices, but will still display the caret for older devices.

ianhanniballake:
@ronaldoeyeem - I've updated my repository (https://github.com/ianhanniballake/ActionBarSherlock) with that fixed. @JakeWharton - would you like me to create a new pull request with these changes and that new commit or wait until we resolve this pull request?

yincrash:
For a temporary solution for the caret if you want to go with @animaonline's (temporary) solution, you can change the up indicator in your Activity's theme with something like

```
    <!-- DrawerToggle can't affect the ABS Up Icon yet, so set it here. -->
    <style name="AppTheme.Main">
        <item name="homeAsUpIndicator">@drawable/ic_drawer</item>
    </style>
```

Of course this won't make the icon slide, but it's a little better graceful degradation than having the up caret.

jameswald:
In honor of Google not yet publishing updated dependencies to public repositories, I've taken this one step further with another implementation.  This one is based on support-v4 r18 source code and includes an ABS demo.

GitHub repo/branch: https://github.com/jameswald/ActionBarSherlock/tree/navigation-drawer
More info from [pull request #2](https://github.com/jameswald/ActionBarSherlock/pull/2):

Provides navigation drawer support back to API 7 on par with the Android 4.3 implementation.  [`setHomeActionContentDescription()`](http://developer.android.com/reference/android/app/ActionBar.html#setHomeActionContentDescription%28java.lang.CharSequence%29) only has an effect on Android 4.3.  This behavior is identical to the official support implementation.

See the ABS demo [`NavigationDrawer.java`](https://github.com/jameswald/ActionBarSherlock/blob/navigation-drawer/actionbarsherlock-samples/demos/src/com/actionbarsherlock/sample/demos/NavigationDrawer.java) for a working example.  In this particular demo, you may notice that the selected item's background will only be highlighted on Honeycomb+.  This is due to the use of the built in [`activatedBackgroundIndicator`](http://developer.android.com/reference/android/R.attr.html#activatedBackgroundIndicator) added in v11.  It may be replaced with a custom drawable to support all platforms.

The Android training guide [Creating a Navigation Drawer](http://developer.android.com/training/implementing-navigation/nav-drawer.html) is still based on r13 and has not yet been updated for r18.  In addition to the training guide, your Activity class must implement [`ActionBarDrawerToggle.DelegateProvider`](http://developer.android.com/reference/android/support/v4/app/ActionBarDrawerToggle.DelegateProvider.html) and should replace the use of [`ActionBarDrawerToggle`](http://developer.android.com/reference/android/support/v4/app/ActionBarDrawerToggle.html) with the convenience classes [`SherlockActionBarDrawerToggle`](https://github.com/jameswald/ActionBarSherlock/blob/navigation-drawer/actionbarsherlock/src/com/actionbarsherlock/app/SherlockActionBarDrawerToggle.java) and [`SherlockActionBarDrawerToggleDelegate`](https://github.com/jameswald/ActionBarSherlock/blob/navigation-drawer/actionbarsherlock/src/com/actionbarsherlock/app/SherlockActionBarDrawerToggleDelegate.java).  Alternatively, you may implement your own toggle and toggle delegate classes.

This cannot be made available for general use because Google has not yet published Android 4.3_r1 or support-v4 r18 to a public Maven repository.  Until that day comes, this change will prevent ABS from building without additional efforts to get the Android 4.3_r1 and support-v4 r18 dependencies installed in your local Maven repository.

JakeWharton:
> Google has not yet published Android 4.3_r1 or support-v4 r18 to a public Maven repository

FYI Google has never published API jars or support library jars to Maven central and will not do so moving forward.

PR #951, opened 2013-05-24
upgrade to support-v4 r13

ceefour:


JakeWharton:
ActionBarSherlock gets deployed to Maven central which means that it can only depend on artifacts that are also available in Maven central. Currently on r7 is deployed which is why the version in this repo is r7 as well.

Someday it will be updated and someday we'll all get to enjoy it out of the box. Until then, you just have to update your copy locally.

PR #943, opened 2013-05-16
Add a Gradle build file

jobi:

PR #952, opened 2013-05-25
Add build.gradle for people moving to Studio

HenriqueRocha:
I'm trying to move to Studio instead of using Eclipse and I had to create this file which will probably be useful for people also moving to Studio and using ABS.
Thanks for your effort in putting together this library.


SimonVT:
Duplicate #943

PR #963, opened 2013-06-06
Improve handling of popup menus and expose PopupMenu

### LINKED ISSUE #752: NullPointerException onScrollChanged()
More info, seems that the bug only appears on Gingerbread and earlier versions.

Its very similar to this issue: https://github.com/JakeWharton/ActionBarSherlock/issues/487

java.lang.NullPointerException
at android.widget.PopupWindow$1.onScrollChanged(PopupWindow.java:127)
at android.view.ViewTreeObserver.dispatchOnScrollChanged(ViewTreeObserver.java:607)
at android.view.ViewRoot.draw(ViewRoot.java:1453)
at android.view.ViewRoot.performTraversals(ViewRoot.java:1335)
at android.view.ViewRoot.handleMessage(ViewRoot.java:1991)
at android.os.Handler.dispatchMessage(Handler.java:99)
at android.os.Looper.loop(Looper.java:150)
at android.app.ActivityThread.main(ActivityThread.java:4385)
at java.lang.reflect.Method.invokeNative(Native Method)
at java.lang.reflect.Method.invoke(Method.java:507)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:849)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:607)
at dalvik.system.NativeStart.main(Native Method)

ksperling:
This should fix #752.
- Work around PopupWindow's leaking of listeners into the ViewRoot VTO
  when the anchor view is detached before the popup is dismissed.
- Apply similar changes to MenuPopupHelper. This also means that support for
  OnAttachStateChangeListener is not required for correct operation of the
  popup window anymore. OnAttachStateChangeListener is still used in case
  the popup anchor view is not attached when the popup is shown, even though
  this seems unlikely to happen in practice as positioning of the popup is
  pretty meaningless in that case. It would be possible to remove the entire
  OnAttachStateChangeListener shim as a second step.
- Included PopupMenu from AOSP. I'm aware this has been previously declined as a feature request as technically it's not part of the ActionBar, However, ABS uses popup menus internally and all the plumbing and resources are already in place, so adding this single class to actually expose what's already there saves a lot of duplication and hassle.

### DIFF
--- a/actionbarsherlock/src/com/actionbarsherlock/internal/view/menu/MenuPopupHelper.java
+++ b/actionbarsherlock/src/com/actionbarsherlock/internal/view/menu/MenuPopupHelper.java
@@ -115,10 +115,20 @@ public class MenuPopupHelper implements AdapterView.OnItemClickListener, View.On
 
         View anchor = mAnchorView;
         if (anchor != null) {
-            final boolean addGlobalListener = mTreeObserver == null;
-            mTreeObserver = anchor.getViewTreeObserver(); // Refresh to latest
-            if (addGlobalListener) mTreeObserver.addOnGlobalLayoutListener(this);
-            ((View_HasStateListenerSupport)anchor).addOnAttachStateChangeListener(this);
+            // Don't attach to the VTO unless the anchor itself is attached to avoid VTO-related leaks.
+            if (anchor.getWindowToken() != null) {
+                ViewTreeObserver vto = anchor.getViewTreeObserver();
+                if (vto != mTreeObserver) {
+                    if (mTreeObserver != null && mTreeObserver.isAlive()) {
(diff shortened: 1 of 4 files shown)


ksperling:
`mvn clean verify` passes for me, the Travis failure seems to be a gradle problem.

ksperling:
See my more detailed comment regarding the underlying problems with ViewTreeObserver: https://github.com/JakeWharton/ActionBarSherlock/issues/752#issuecomment-19018122

JakeWharton:
I merged this but removed `PopupMenu`. If you need this class it can be included in the downstream app and use the internal ABS classes at the risk of them changing behavior or coming/going at any point in the future.

PR #964, opened 2013-06-06
Download build tools package as part of CI.

JakeWharton:


JakeWharton:
Took more than 5 min because of SDK ridiculousness. Oh well.

Prototik:
@JakeWharton build tools available only since r22 asdk, but for travis ci we cannot use r22 because it wants to accept some licenses from standard console input. Veeery funny. I don't know how to convince Google that we agree this shit. Ops, sorry, this license.

PR #974, opened 2013-06-19
Travis. ASDK r22 and correct gradle build.

Prototik:
Yea, i know - holy shit in travis config...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,14 +6,70 @@ branches:
 
 notifications:
   email: false
+  
+env:
+  global:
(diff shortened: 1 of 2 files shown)


Prototik:
@JakeWharton fu*k, yes, i know! :D
But standard way for installing doesn't work for r22 - it's asks about license.

Of course you can write to asdk devs for providing something more sexy :)

JakeWharton:
Needs a quick rebase. Should only be one conflict in the root build.gradle that's easily resolved (use the version from dev)

PR #980, opened 2013-06-27
Fix missing android namespace prefix

bbcallen:
From lint check


JakeWharton:
The platform search resources have this which is why it's echoed in ABS. The fact that it's wrong mirrors the behavior of the native action bar search view.

PR #985, opened 2013-07-03
Add ActionBarDrawerToggle which compatible with ActionBarSherlock

TangKe:
device-2013-07-03-112721
Just like the Google Plus.


JakeWharton:
I'll merge this when the support library in Maven central gets update to r13. Thanks for the contribution.

PR #994, opened 2013-07-15
update android build tools to 0.5.+

googolmo:
Android Gradle plug-in 0.5.0. This new version of the plug-in is not backwards compatible. When opening a project that uses an older version of the plug-in, Studio will show the following error as a balloon in the upper right corner of the IDE:

You will need to change the version of the Android Gradle plug-in to 0.5.0. Just click the link “Search in build.gradle files” to look for the places where this change needs to be made. (See the Troubleshooting section below for more info).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ buildscript {
   }
 
   dependencies {
-    classpath 'com.android.tools.build:gradle:0.4.+'
+    classpath 'com.android.tools.build:gradle:0.5.+'
   }
 }
 
@@ -20,5 +20,7 @@ allprojects {
     options.encoding = "UTF-8"
   }
 }
+
+task assemble{}
 
 apply plugin: 'android-reporting'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -21,4 +21,6 @@ allprojects {
   }
 }
 
+task assemble{}
+
```
JakeWharton: Why?

googolmo: i think this it's a bug,
if remove it gradle build will failed

JakeWharton: In Android Studio or on the command line?

googolmo: Android Studio

JakeWharton: I think you just need to re-import the project from the `build.gradle` file and delete the existing project when it prompts.

googolmo: ok i will try it 
and i have deleted it

googolmo:
@JakeWharton i sync project with gradle file and it show failed
i think this is android studio 2.0 's bug

PR #996, opened 2013-07-20
Update gradle build files so that ABS can be referenced correctly as a gradle subproject

ghost:
Hi I have updated the gradle build files so that anyone can correctly reference ABS as a gradle subproject.
I followed Xavier's recommendation found here: http://stackoverflow.com/a/16974411.
This means that I have updated settings.gradle as recommended and I also updated each project's build.gradle so that each subproject applies both the buildscript repository and classpath dependency individually. 
Finally I changed the dependency for the support-v4.jar to v13.0+. I know that is not yet available in Maven Central but if someone is building locally I believe the latest version should be used.

### DIFF
--- a/actionbarsherlock-i18n/build.gradle
+++ b/actionbarsherlock-i18n/build.gradle
@@ -1,3 +1,13 @@
+buildscript {
+  repositories {
+    mavenCentral()
+  }
+
+  dependencies {
+    classpath 'com.android.tools.build:gradle:0.5.+'
+  }
+}
+
 apply plugin: 'android-library'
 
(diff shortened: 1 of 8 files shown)


schmidp:
Could you please add a comment on how to use ABS with your changes as a subproject?

I can't manage to get my settings.gradle and build.gradle setup correctly.

iNoles:
@schmidp Did you installed Android Support Repo from the SDK Manager?

JakeWharton:
This won't be needed. It will be in Maven central soon.

PR #1000, opened 2013-07-28
fixed "textAppearance" name conflict

sephiroth74:


JakeWharton:
Conflict with what?

sephiroth74:
it will conflict with most of existing user applications. "textAppearance" is a very common attribute name. Is there a particular reason why all the attributes dont have a prefix like "abs_" ?

SimonVT:
They're not prefixed so the only difference between the native and ABS styles is the android: prefix. ABS defines all attributes so it's not a guessing game whether it should be prefixed or not.
There's no reason you couldn't just use android:textAppearanceSmall in your own apps, or the ABS-defined textAppearanceSmall.
Besides, it's too late in the game to change this anyway imo. And if it should be done, it should be done consistently.
Also, please don't update the support library.

PR #1004, opened 2013-07-31
make performItemAction compatible with ActionProvider sub menu

ol-v-er:
In the case of the submenu of an ActionItem is provided by an ActionProvider and not directly by the ActionItem, the submenu was not displayed onClick. 

It was working with native implementation and not with ActionBarSherlock one. 

I've backport the method from Android code source to have the correct implementation.

### DIFF
--- a/actionbarsherlock/src/com/actionbarsherlock/internal/view/menu/MenuBuilder.java
+++ b/actionbarsherlock/src/com/actionbarsherlock/internal/view/menu/MenuBuilder.java
@@ -884,19 +884,26 @@ public class MenuBuilder implements Menu {
 
         boolean invoked = itemImpl.invoke();
 
+        final ActionProvider provider = item.getActionProvider();
+        final boolean providerHasSubMenu = provider != null && provider.hasSubMenu();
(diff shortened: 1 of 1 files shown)


SimonVT:
The ABS compat implementation follows the behavior of ICS. I don't see this change when checking out ICS, does it work there?

ol-v-er:
It looks like an Android bug.

In Android 4.0.4, the code source of the performItemAction method is the same as the one in ABS.
In Android 4.1, they fixed it as the one in the pull request.

The commit related on Android code source:

```
commit f77f480800a84ceb377e47cc200baf2bae4f5d9a
Author: Adam Powell <adamp@google.com>
Date:   Mon May 14 16:44:43 2012 -0700

Fix ActionProviders that have submenus

Automatically create a submenu for items with ActionProviders that
provide a submenu when the item was not initially configured to have
one.

Bug 6493376

Change-Id: I27218e2ab205e87534f235eb04042b185d99d143
```

SimonVT:
If ABS behaves the same as ICS then we can't change it. You'll have to come up with another approach in your app, or drop support for <api16 :p

PR #1006, opened 2013-08-03
Fix cell usage calculation for menu items with text

goncalossilva:
This prevents an edge case where menu items with text have 1 available cell and end up using it, thus overlapping the overflow button.

This is based on Android 4.2.2's code for ActionMenuView: http://grepcode.com/file/repository.grepcode.com/java/ext/com.google.android/android/4.2.2_r1/com/android/internal/view/menu/ActionMenuView.java#ActionMenuView.measureChildForCells%28android.view.View%2Cint%2Cint%2Cint%2Cint%29


SimonVT:
Pretty sure this is a framework bug on ICS. We can't fix this in ABS.

goncalossilva:
Yes, it was a bug in ICS which was later fixed.

Why can't you apply the JB fix to ABS? I mean, right now ABS is consistent with ICS (buggy). After applying the patch it would be consistent with JB (fixed). What's the catch?

SimonVT:
Because the native action bar is used on ICS. You have to work around this in your app anyway, so for easier diacoverability we decided to make it consistent with ICS - including bugs.

goncalossilva:
:+1:

PR #1020, opened 2013-08-22
Move makefile and classpath

matthewmichihara:
The lib dir changed from `library` to `actionbarsherlock` in ABS 4.3


matthewmichihara:
Oops sorry wrong repo.

PR #1025, opened 2013-08-28
update to sdk 18 + remove warnings

hielkehoeve:
- updated to sdk 18 + updated support library
- removed unsused private functions
- fixed deprecated functions by wrapping them in functions that perform an api check


SimonVT:
ABS is compiled with API16 (but project.properties isn't used with maven anyway), the support library can't be updated since it's not in central.
There's no reason for the rest of the changes, they do nothing but make the codebase more complex.

PR #1038, opened 2013-09-13
update usage docs to mention gradle.

mlc:
also there was a list with four things in it where the introduction said there were three of them.


JakeWharton:
Merged to `dev` and deployed.

Issue #652, opened 2012-10-11
got NullPointerException on SherlockFragmentActivity.invalidateOptionsMenu


kingori:
I'm using ABS 4.2.0 on GB(2.3.6) and got NullPointerException.

Stacktrace is like below. I've put 3 fragment on each tab and each tab has its own action menu item.

10-11 14:03:19.566: ERROR/AndroidRuntime(25214): FATAL EXCEPTION: main
        java.lang.NullPointerException
        at android.view.View.dispatchRestoreInstanceState(View.java:6326)
        at android.view.ViewGroup.dispatchRestoreInstanceState(ViewGroup.java:1211)
        at android.view.View.restoreHierarchyState(View.java:6310)
        at com.actionbarsherlock.internal.view.menu.MenuBuilder.restoreActionViewStates(MenuBuilder.java:361)
        at com.actionbarsherlock.internal.ActionBarSherlockCompat.preparePanel(ActionBarSherlockCompat.java:503)
        at com.actionbarsherlock.internal.ActionBarSherlockCompat.dispatchInvalidateOptionsMenu(ActionBarSherlockCompat.java:272)
        at com.actionbarsherlock.app.SherlockFragmentActivity.invalidateOptionsMenu(SherlockFragmentActivity.java:150)
        at com.actionbarsherlock.app.SherlockFragmentActivity.supportInvalidateOptionsMenu(SherlockFragmentActivity.java:156)
        at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1095)
        at android.support.v4.app.BackStackRecord.run(BackStackRecord.java:635)
        at android.support.v4.app.FragmentManagerImpl.execPendingActions(FragmentManager.java:1431)
        at android.support.v4.app.FragmentManagerImpl$1.run(FragmentManager.java:420)
        at android.os.Handler.handleCallback(Handler.java:587)
        at android.os.Handler.dispatchMessage(Handler.java:92)
        at android.os.Looper.loop(Looper.java:130)
        at android.app.ActivityThread.main(ActivityThread.java:3687)
        at java.lang.reflect.Method.invokeNative(Native Method)
        at java.lang.reflect.Method.invoke(Method.java:507)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:847)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:605)
        at dalvik.system.NativeStart.main(Native Method)

[...]

PR #1042, opened 2013-09-20
Fix issue #652 NPE on all sdks

mazzola:
Calling dispatchRestoreInstanceState from ABS causes View.java to throw a NPE. This bug was appearing on devices running 3.1, 3.2, and 3.2.1

View.java doesn't check for a null container in dispatchRestoreInstanceState so we need to check it here.


JakeWharton:
Dupe of #1044.

PR #1056, opened 2013-10-15
Implement a left margin on the home icon

fomojola:
Trying to solve the use case where the action bar up is not enabled but the home icon is centered. Generally this requires creating a custom home layout and overriding (otherwise the left margin isn't quite right), but this allows that margin to be overriden by using a custom style like this:

```
<?xml version="1.0" encoding="utf-8"?>
<resources>
    <style name="Theme.Styled" parent="Theme.Sherlock.Light.DarkActionBar">
        <item name="actionBarStyle">@style/Widget.Styled.ActionBar</item>
        <item name="android:actionBarStyle">@style/Widget.Styled.ActionBar</item>
        <item name="homeAsUpIndicator">@null</item>
        <item name="android:homeAsUpIndicator">@null</item>
    </style>

    <style name="Widget.Styled.ActionBar" parent="Widget.Sherlock.Light.ActionBar.Solid.Inverse">
        <item name="homeAsUpIndicator">@null</item>
        <item name="android:homeAsUpIndicator">@null</item>
    </style>

    <dimen name="abs__action_bar_icon_left_margin">12dip</dimen>
</resources>
```

### DIFF
--- a/actionbarsherlock/res/layout/abs__action_bar_home.xml
+++ b/actionbarsherlock/res/layout/abs__action_bar_home.xml
@@ -29,6 +29,7 @@
     <ImageView android:id="@id/abs__home"
                android:layout_width="wrap_content"
                android:layout_height="wrap_content"
+               android:layout_marginLeft="@dimen/abs__action_bar_icon_left_margin"
                android:layout_marginRight="8dip"
                android:layout_marginTop="@dimen/abs__action_bar_icon_vertical_padding"
                android:layout_marginBottom="@dimen/abs__action_bar_icon_vertical_padding"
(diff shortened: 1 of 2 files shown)


JakeWharton:
Not only should you never override a dimension by convention, but this is something that the native action bar does not allow you to do. How are you supporting Android 4.0+?

fomojola:
Ah, actually I am: it renders fine on 4.0+ but not on 2.2. Didn't realize that overriding the dimension was verboten: this was explicitly a way to control the margins on <4.0, since I saw many many many attempts to adjust the left margin that simply didn't work when deployed, but this one did. Its in my private copy of ActionBarSherlock: I'd thought to share it but if there's a better way to make that margin adjustable I'd love to get it in: for something that small it seems particularly complex to require a custom home layout.

JakeWharton:
ABS only offers the API that's available on Android 4.0. Since there's nothing to control the margin in native Android we can't add it to this library. I realize that it inhibits customizability, but the point of this library was not to facilitate a customizable action bar but rather to bring the 4.0 API backwards through all the old APIs one would need to support.

PR #1061, opened 2013-11-01
Update abs__strings.xml (SK)

pylerSM:


Prototik:
i18n module grabs strings from AOSP, i don't think that you should modify it.
Although if Jake allow it - i can show you how to correct override strings for resbuilder plugin.

JakeWharton:
We maintain the translations in accordance with Android 4.0.x since that's the version of the action bar that ABS provides.

PR #1074, opened 2013-12-12
Force OverflowButton to show, to conform to the expected behaviour in th...

lionzeye:
...e latest Android Framework

Here 's why ;)
https://android.googlesource.com/platform/frameworks/base.git/+/ea04f3cfc6e245fb415fd352ed0048cd940a46fe


SimonVT:
The expected behavior pre-kitkat is that the overflow item doesn't show.

PR #1081, opened 2014-01-09
No more Android Support Library mismatch errors

rawkode:
Removing the Android Support Library would allow people to integrate int...o their projects easier and not have to consistently update the library themselves to match their project version. Lets allow people to drop their copy in once.


RuellBrown:
:+1:

JakeWharton:
Breaks Eclipse devs so can't do that. Blame Eclipse and/or ADT.

PR #1088, opened 2014-02-12
travis build status pic

paulvi:
I ran into #1087 , showing current build status would be better

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 ActionBarSherlock
 =================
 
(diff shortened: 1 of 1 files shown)


JakeWharton:
I find absolutely no utility in these things. It should always be green, if not it should be emailing people to annoy them.

paulvi:
You should have learned basic Travis configuration.

JakeWharton:
Ya I'll enroll in a Coursera for it.

